PR #16, opened 2022-07-22
Create cargo workspace

ssnover:
I'm planning a proc-macro for helping with message generation and it will need to be in a separate crate. I preemptively have made a workspace now in order to prevent future merge conflicts due to it.


ssnover:
By the way, is there any reason you're not include `Cargo.lock`?

Carter12s:
No reason on cargo.lock,

PR #19, opened 2022-08-19
General Clean-ups of Documentation and Public API ahead of first release

Carter12s:


roslibrust/src/comm.rs
```diff
@@ -76,7 +80,7 @@ impl FromStr for Ops {
 
 /// Describes the low level comm capabilities of talking to a rosbridge server
 #[async_trait]
-pub trait RosBridgeComm {
+pub(crate) trait RosBridgeComm {
```
ssnover, line 83: Is the intention to not allow external code to impl this trait?

ssnover (approved):
All of the comments look good. Unfortunately, many of them are on codegen which I've just finished rewriting :smile:

ssnover:
One thing I'd advocate for (maybe prior to publishing on crates.io) is to recreate the ROS tutorial in a mdbook and then host it with Jekyll from a GitHub repo. This will make it easier to show how to get started.

roslibrust/src/comm.rs
Carter12s: It is "for now" I don't think anyone should be impl-ing until we have a better plan going forward.

Carter12s:
> One thing I'd advocate for (maybe prior to publishing on crates.io) is to recreate the ROS tutorial in a mdbook and then host it with Jekyll from a GitHub repo. This will make it easier to show how to get started.

Filed https://github.com/Carter12s/roslibrust/issues/22

PR #21, opened 2022-08-21
Generate messages supporting macro workflow and tooling workflow

ssnover:
I'm so sorry for the size of this.

Currently all of the source compiles, but one example is broken and I want to get some input from you @Carter12s on how best to proceed on that. There weren't any other service files at the time and the pattern used in the `calling_service` example looks highly irregular in how it defines its own source.

Seems like maybe the examples should be done like in [`tonic`](https://github.com/hyperium/tonic/tree/master/examples) where they're not typical examples since setting one up is a little more involved; requiring a rosbridge and multiple processes.

Messages can be generated into a file with the binary defined in `roslibrust_test` by running `cargo run --bin roslibrust_test > output_lib.rs`. This also demonstrates using the function to generate code.

The macro can be used by invoking it with no arguments, or with a comma-separated list of paths, as seen under `roslibrust_test/src/lib.rs`.

This MR introduces a few new crates to the workspace. The one containing the proc-macro is necessary due to how proc-macros work. The others add deployment overhead, but make it more convenient for the user of the `roslibrust` crate since they'll be compiling less code if they're not regenerating messages all of the time.

A possible improvement to discuss is wrapping all of the generated structs in modules to reflect the package they came from. If you generated in lib.rs in a package called `ros_msgs` like this:
```rust
use roslibrust_codegen_macro::find_and_generate_ros_messages;
find_and_generate_ros_messages!();
```
You currently access them as a client like `ros_msgs::Float64`. If the package's module name was introduced, it would be `ros_msgs::std_msgs::Float64` in which case you could do `use ros_msgs::*` at the top and all accesses would look just like C++.


roslibrust_codegen/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
```
Carter12s, line 1: Since this will be published as a separate crate, needs LICENSE, README.md, extra fields under package

roslibrust/Cargo.toml
```diff
@@ -31,6 +31,7 @@ thiserror = "1.0.31"
 [dev-dependencies]
 # Used for logging in examples
 simple_logger = "2.1.0"
+roslibrust_test = { path = "../roslibrust_test" }
```
Carter12s, line 34: [nit]

I'm starting to be a really big fan of adding comments to Cargo.toml to explain why dependencies exist.
Would generally like a comment here like "contains test message definitions for use in examples and integration tests"

roslibrust_codegen/src/lib.rs
```diff
@@ -0,0 +1,230 @@
+use proc_macro2::TokenStream;
+use quote::{format_ident, quote};
+use roslibrust_rospack as rospack;
```
Carter12s, line 3: I'm a little leery of using the rospack name given that rospack does a bunch of things we don't do, and I don't want to imply that were publishing a crate that will eventually support all rospack functionality.

Maybe this is okay, but gut reaction was roslibrust_util could be a better name...

Also I'm starting to look ahead to what ros2 support will look like and not sure if we would need a ros1pack vs ros2pack...

roslibrust_codegen/src/lib.rs
```diff
@@ -0,0 +1,230 @@
...
+pub fn find_and_generate_ros_messages(additional_search_paths: Vec<PathBuf>) -> TokenStream {
+    let mut search_paths = rospack::get_search_paths();
+    search_paths.extend(additional_search_paths.into_iter());
+    let mut packages = rospack::crawl(search_paths);
+    packages.dedup_by(|a, b| a.name == b.name);
+    let mut message_files = packages
+        .iter()
+        .map(|pkg| {
+            rospack::get_message_files(pkg)
+                .unwrap()
```
Carter12s, line 21: Should be an expect with a more clear error message "Failed to get message files for pkg XXX due to XXX"

roslibrust_codegen/src/lib.rs
```diff
@@ -0,0 +1,230 @@
...
+                .into_iter()
+                .map(|path| (pkg.name.clone(), path))
+        })
+        .flatten()
+        .collect::<Vec<_>>();
+    let service_files = packages
+        .iter()
+        .map(|pkg| {
+            rospack::get_service_files(pkg)
+                .unwrap()
```
Carter12s, line 31: Should be an expect with a more clear error message

roslibrust_codegen/src/lib.rs
```diff
@@ -0,0 +1,230 @@
...
+    {
+        if seen_count > MAX_PARSE_ITER_LIMIT {
+            parsed.fields.iter().for_each(|field| {
+                if ROS_TYPE_TO_RUST_TYPE_MAP.contains_key(field.field_type.field_type.as_str()) {
+                    println!("RESOLVED PRIMITIVE: {}", field.field_type.field_type);
+                } else if message_map.contains_key(&field.field_type.field_type) {
+                    println!("RESOLVED MSG: {}", field.field_type.field_type);
+                } else {
+                    println!("UNRESOLVED: {}", field.field_type.field_type);
+                }
```
Carter12s, line 119: eprintln! ?

Probably want a "MAX_PARSE_ITER_LIMIT Exceeded, likely circular dependency message?"

roslibrust_codegen_macro/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
```
Carter12s, line 1: LICENSE, README, extra fields here

roslibrust_rospack/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
```
Carter12s, line 1: README, LICENSE, extra fields

roslibrust/Cargo.toml
```diff
@@ -31,6 +31,7 @@ thiserror = "1.0.31"
 [dev-dependencies]
```
Carter12s, line 31: Can we change edition to 2021 above here, so we have consistency across workspace

roslibrust_rospack/src/lib.rs
```diff
@@ -0,0 +1,147 @@
...
+        eprintln!("No ROS_PACKAGE_PATH defined.");
+        vec![]
+    }
+}
+
+pub fn crawl(search_paths: Vec<PathBuf>) -> Vec<Package> {
+    let mut packages = vec![];
+
+    for path in search_paths {
+        if let Ok(found_packages) = packages_from_path(path, 1000) {
```
Carter12s, line 38: Named constant?

roslibrust_rospack/src/lib.rs
```diff
@@ -0,0 +1,147 @@
...
+            }
+        }
+    } else {
+        eprintln!("{} is not a directory", path.to_string_lossy())
+    }
+
+    Ok(found_packages)
+}
+
+fn is_metapackage(package_xml_path: &Path) -> io::Result<bool> {
```
Carter12s, line 100: Is there a test for this? Not super familiar with these but seems slightly wonky that we would need to recognize

roslibrust_rospack/test_package/package.xml
```diff
@@ -0,0 +1,21 @@
+<package>
+  <name>test_msgs</name>
+  <version>0.1.0</version>
+  <description>
+    This is a test package with message definitions.
+  </description>
+  <maintainer email="ssnover95@gmail.com">ssnover</maintainer>
+  <license>None</license>
```
Carter12s, line 8: MIT

roslibrust_test/src/lib.rs
```diff
@@ -0,0 +1,3 @@
+use roslibrust_codegen_macro::find_and_generate_ros_messages;
+
+find_and_generate_ros_messages!("..");
```
Carter12s, line 3: I'm curious about how often this gets evaluated / if we prefer to have a static version here.

roslibrust_test/src/main.rs
```diff
@@ -0,0 +1,34 @@
+use std::borrow::Cow;
+use std::io::Write;
+use std::process::{Command, Stdio};
+
+fn main() {
+    let tokens = roslibrust_codegen::find_and_generate_ros_messages(vec!["..".into()]);
+    let source = format_rust_source(tokens.to_string().as_str()).to_string();
+    println!("{}", source);
+}
```
Carter12s, line 9: Should this just be a test instead of main?

roslibrust_codegen/src/lib.rs
ssnover: I don't think we necessarily need to support everything that `rospack` currently does. But `rospack` does a lot of things that will be useful and using via shelling out or binding the C++ library will be a huge nuisance. If I was using `roslibrust` and needed to do something that `rospack` does, naming it explicitly that seems like it'd help me find it.

On the other hand, it is a little awkward if we define some functions that aren't done by `rospack`. Not sure where I fall on it.

roslibrust_codegen/src/lib.rs
ssnover: Ah I actually intended to take this out. There's a message directly below for that.

roslibrust_rospack/src/lib.rs
ssnover: I believe I added this because it broke my initial implementation, but I can't actually find any examples that would cause that (and the spec for metapackages seems to indicate they shouldn't). I'll remove all the metapackage stuff.

roslibrust_test/src/lib.rs
ssnover: Hmm good point that I hadn't considered. There's not really any way for `cargo` to know this crate has dependencies on the message files and should re-compile the crate if one of those message files changes. This is a possible footgun. I'm not sure if that warrants taking it out or if it should just get a lot of explicit documentation around when it will/won't recompile.

roslibrust_test/src/main.rs
ssnover: This package should probably be renamed. My intention is that this is an internal tool, not published to crates.io which can be used to generate code for the `local_msgs` in this repo. So the lib will maintain the generated code and this bin will act as both a generator and documentation on how to write one. Does that make sense?

roslibrust_rospack/test_package/package.xml
ssnover: Eh, it's a test package, but sure.

ssnover:
Funny thing popped up in the example I was fixing which leads me to think we'll need to do the modules inside of the macro. A ROS message `Path.msg` was generated and then caused a multiple definition error because the file invoking the generation macro had `use std::path::Path;` above it. The error was pretty annoying to find with the macro-generated code.

ssnover:
Funny thing popped up in the example I was fixing which leads me to think we'll need to do the modules inside of the macro. A ROS message `Path.msg` was generated and then caused a multiple definition error because the file invoking the generation macro had `use std::path::Path;` above it. The error was pretty annoying to find with the macro-generated code so I went ahead and did it.

Very annoying complications emerged:
* Referring to types defined in the same module requires `self::` prefix.
* Referring to types defined in other modules in the same file requires `use super::pkg_name;`
* Difficult to track package dependencies, so I opted for `#[allow(unused_imports)]`

roslibrust_codegen/src/lib.rs
Carter12s: Good enough for now!

roslibrust_test/src/lib.rs
```diff
@@ -0,0 +1,1885 @@
+#[allow(unused_imports)]
```
Carter12s, line 1: fmt... I think I'll try to tackle that as a way to test these changes out.

Carter12s: Okay so I did some digging and I hate things...

I personally have some major style gripes with this:

1. There aren't empty lines between blocks
2. the ":: serde :: Deserialize" feels like it should really be "::sered::Deserialize"

However neither of these appears fixable with a .rustfmt file so I'm going to have to live with it. I am at least adding a lint check to CI cause we don't have one.

Carter12s:
So slightly interesting / odd behavior for testing. I'm compiling on a system without ROS_PACKAGE_PATH defined.

```
$ cargo test
No ROS_PACKAGE_PATH defined.
No ROS_PACKAGE_PATH defined.
No ROS_PACKAGE_PATH defined.
    Finished test [unoptimized + debuginfo] target(s) in 0.03s
     Running unittests (/home/carter/open_source/roslibrust/target/debug/deps/roslibrust-285f99fbff3bc59e)

running 1 test
test util::tests::get_installed_msgs_test ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

   Doc-tests roslibrust

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
```

The eprintln!() for No ROS_PACKAGE_PATH defined is printing every time (and it also prints it in CI)... We may want to turn that into a warn!() call, but I'm not 100% sure about that... Current behavior feels wrong...

Carter12s:
Will also say that at this point I'm pretty happy to merge this MR and deal with follow up in other isseues. Actually I'm just going to do that.

PR #28, opened 2022-08-31
Support service server

Carter12s:
Adds the ability to run a service server.

This is pretty basic so far and pretty minimally tested, needs more work but should be a good start

### DIFF
--- a/roslibrust/src/integration_tests.rs
+++ b/roslibrust/src/integration_tests.rs
@@ -1,11 +1,11 @@
-/// Tests in this module require a running rosbridge server supportin websockets on port 9090
+/// Tests in this module require a running rosbridge server supporting websocket connections on port 9090
 /// run `docker-compose -f ros1_bridge_compose.yaml up` in the docker folder of this repo to start a bridge
 /// and then run `cargo test --features running bridge` to execute these tests
 /// NOTE: these aren't stored in "tests" dir as many tests need access to private functions like publish()
 #[cfg(test)]
 #[cfg(feature = "running_bridge")]
 mod integration_tests {
-    use crate::{Client, ClientOptions, RosMessageType, Subscriber};
+    use crate::{Client, ClientOptions, RosMessageType, RosServiceType, Subscriber};
     use log::debug;
     use tokio::time::{timeout, Duration};
     // On my laptop test was ~90% reliable at 10ms
@@ -34,6 +34,31 @@ mod integration_tests {
     }
 
     type TestResult = Result<(), anyhow::Error>;
+
+    // Note: have to use a real .srv which exists in our docker image
+    pub struct SetBoolSrv {}
+    impl RosServiceType for SetBoolSrv {
+        const ROS_SERVICE_NAME: &'static str = "std_srvs/SetBool";
+        type Request = SetBoolRequest;
+        type Response = SetBoolResponse;
+    }
+
+    #[derive(serde::Deserialize, serde::Serialize, Debug, Default, Clone, PartialEq)]
+    pub struct SetBoolRequest {
+        data: bool,
+    }
+    impl RosMessageType for SetBoolRequest {
+        const ROS_TYPE_NAME: &'static str = "std_srvs/SetBool";
+    }
+
+    #[derive(serde::Deserialize, serde::Serialize, Debug, Default, Clone, PartialEq)]
+    pub struct SetBoolResponse {
+        success: bool,
+        message: String,
+    }
+    impl RosMessageType for SetBoolResponse {
+        const ROS_TYPE_NAME: &'static str = "std_srvs/SetBool";
+    }
 
     /**
     This test does a round trip publish subscribe for real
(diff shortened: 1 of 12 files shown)


roslibrust/src/integration_tests.rs
```diff
@@ -205,4 +230,42 @@ mod integration_tests {
...
+            |_req: SetBoolRequest| -> Result<SetBoolResponse, Box<dyn std::error::Error + Send + Sync>> {
+                Ok(SetBoolResponse {
+                    success: true,
+                    message: "call_success".to_string(),
+                })
+            };
+
+        let topic = "/self_service_call";
+
+        client
```
ssnover, line 254: Would it make sense for the API to return a handle such that the drop of said handle would unadvertise the service like in roscpp?

roslibrust/src/lib.rs
```diff
@@ -338,16 +361,44 @@ impl Client {
...
+            let response_string = serde_json::json!(response);
+            Ok(response_string)
+        };
+
+        let res = self
+            .services
+            .insert(topic.to_string(), Box::new(erased_closure));
+        if let Some(_previous_server) = res {
+            warn!("Re-registering a server for a pre-existing topic? Are you sure you want to do this");
+            unimplemented!()
```
ssnover, line 395: Lol I hope not because this will crash.

roslibrust/src/lib.rs
```diff
@@ -338,16 +361,44 @@ impl Client {
...
+        }
+
+        //TODO need to provide unadvertise and drop callback
```
ssnover, line 398: Ah disregard my other comment about Drop then. Perhaps an issue to make sure this doesn't stay to-do (or someone can claim it).

roslibrust_codegen/src/lib.rs
```diff
@@ -91,26 +108,37 @@ struct MessageMetadata {
...
-                        parsed_messages.push_back(MessageMetadata {
-                            package: pkg.clone(),
-                            path: path.clone(),
-                            seen_count: 0,
-                            parsed,
-                            unparsed,
-                        });
-                    });
+                let srv_file = parse_ros_service_file(contents, name.to_string(), &pkg);
+                // TODO stop cloning with reckless abandon
```
ssnover, line 126: Lol good luck with this. All of these need to outlive the scope where we read them from the file. They're just strings ¯\_(ツ)_/¯

roslibrust_codegen/src/lib.rs
```diff
@@ -127,7 +155,11 @@ fn parse_message_files(
             }
         }
     }
-    Ok(resolve_message_dependencies(parsed_messages))
+    parsed_services.sort_by(|a, b| a.name.cmp(&b.name));
```
ssnover, line 158: Making sure I understand the reasoning: sorting here so that they come out alphabetized in the final source output right?

roslibrust_codegen/src/lib.rs
```diff
@@ -198,6 +230,26 @@ fn derive_attrs() -> Vec<syn::Attribute> {
...
+/// The service definition defines a struct representing the service an an implementation
+/// of the RosServiceType trait for that struct
+fn generate_service(service: ServiceFile) -> TokenStream {
+    let service_type_name = format!("{}/{}", &service.package, &service.name);
+    let struct_name = format_ident!("{}", service.name);
+    let request_name = format_ident!("{}", service.request_type.name);
+    let response_name = format_ident!("{}", service.response_type.name);
+    quote! {
+        pub struct #struct_name {
+
```
ssnover, line 243: I suppose services can't have constants? Or perhaps this is a sufficiently small issue if they can given that someone can get it from the message type that's generated.

roslibrust_test/src/main.rs
```diff
@@ -32,3 +35,27 @@ fn format_rust_source(source: &str) -> Cow<'_, str> {
     }
     Cow::Borrowed(source)
 }
+
+#[cfg(test)]
+mod test {
```
ssnover, line 40: Oh! Big fan of this test.

ssnover:
These changes look like a good start and I think this can be merged. Can you make issues for unimplemented parts (todos) and I'll pick one up?

ssnover (approved):

roslibrust_codegen/src/lib.rs
Carter12s: Yup I wrote an CI test to catch that generation had been re-run after I forgot to do it, but then the CI test failed because the output came out differently ordered on my system vs CI.

roslibrust_codegen/src/lib.rs
Carter12s: Constants within a service will end up existing on either their request or response type, no constants live at the "service scope"

roslibrust_codegen/src/lib.rs
Carter12s: yeah I do think between the services and the message files there is a lot of extra redundant cloning happening as well thou.

roslibrust/src/lib.rs
Carter12s: yeah let me do that, may tackle drop and unadvertise support tomorrow

roslibrust/src/integration_tests.rs
Carter12s: https://github.com/Carter12s/roslibrust/issues/29

roslibrust/src/lib.rs
Carter12s: https://github.com/Carter12s/roslibrust/issues/29

Issue #31, opened 2022-09-01
Add default features for rosbridge_client


ssnover:
Would be good to be able to compile the generated types without necessarily needing to compile the whole crate and all of the crate's dependencies.

PR #32, opened 2022-09-02
Move files around and add features to roslibrust

ssnover:
Since you were asking about which crates were necessary earlier, I took a stab at reducing things down and then reorganizing the `roslibrust` crate around features.

This MR is probably going to be a mess to review, but I promise I didn't change any code, just moved it around. Probably easier to pull and view in editor.

Highlights:
* Added the following features to `roslibrust`: `client`, `utils`, `codegen`, `all`, `default`
* Put all of the rosbridge communication stuff inside a module, but re-exported so it'd be the same.
* Removed need for `roslibrust_codegen` and `roslibrust_rospack`, unsure how to mark them as deprecated on `crates.io`
* Moved message files out of `roslibrust` since we don't need them inside the crate dir structure anymore.
* Deleted the old code in `utils.rs` since it was hanging around from the previous codegen method.

We may also want to consider moving more of the dependencies to be bind the new feature flags, I just did the obviously expensive ones to start.


ssnover:
Closes #31

Carter12s:
I went ahead and merged figuring this was moving in the right direction, and it would be easiest to address stuff by building on top of. Thanks for the clean-up. I'm not sure I can fully remove the dead crates, but I can at least yank them.

Issue #20, opened 2022-08-19
Make Publisher and Subscriber clone()


Carter12s:
Could be useful, current design should support it pretty well.

Carter12s:
I've decided to not do this for now as I think it over complicates writing backends.

Publishers should always be created from .advertise().

PR #35, opened 2022-09-04
Fix bridge client lifetime in spin context

ssnover:
I was looking to tackle the issue where the `Client::stubborn_spin` function would hold onto a Client forever and it turned into a rabbit hole. I ended up moving all of the members to a `ClientInner` struct which is co-owned through `Arc<RwLock<T>>` by the spin context and by all `Client`s.

In the nominal case, the lock is only being accessed as a reader because all of the members themselves are only used immutably (at least directly). This should mean that all accesses are very fast. However, if it disconnects from the bridge, the spin context takes a write lock which will be exclusive and allows it to overwrite the transport. 

This is work in progress right now. One remaining task is there needs to be protection against starving of the write lock since it's totally possible for a user who relies on the stubborn reconnect to spam `Client` methods and for the write lock to be unobtainable.


roslibrust/src/rosbridge/publisher.rs
```diff
@@ -12,6 +12,7 @@ use crate::{Client, RosLibRustResult, RosMessageType};
 ///   - Ability for publish to by const
 // Instead of giving back a publisher should we give back a reference to one, and give back the
 // same reference when you advertise multiple times? Would require non-mut references?
+#[derive(Clone)]
```
Carter12s, line 15: Uhh... I think we can't do this...

This will break unadvertise logic? If we want publisher to be clone we need to keep track of the count of outstanding publishers for a topic and only unadvertise the topic when the count drops to zero. There is a #20 issue to do this.

roslibrust/src/rosbridge/mod.rs
```diff
@@ -94,64 +94,41 @@ impl ClientOptions {
     }
 }
 
-/// A client connection to the rosbridge_server that allows for publishing and subscribing to topics
 #[derive(Clone)]
 pub struct Client {
```
Carter12s, line 98: Seperate todo and probably worth a new issue, but client should be in its own file

Carter12s:
The diff for this one is a mess, but in principle I like the changes.  I might name things slightly differently like ClientHandle and Client vs. Client and ClientInner.

With how messed up the diff is pretty hard to tell what state this is really in, but let me know when its not WIP and I'll take a closer look.

roslibrust/src/rosbridge/publisher.rs
ssnover: Ah good call, will remove.

ssnover:
Main remaining item I want to address is prevent soft deadlocks by users repeatedly retrying methods while waiting for the reconnect and the reconnect being starved of the write lock.

Carter12s:
> Main remaining item I want to address is prevent soft deadlocks by users repeatedly retrying methods while waiting for the reconnect and the reconnect being starved of the write lock.

This is an awesome thing to have an integration test around. I think beyond features, the most important thing we can do to make this crate have value is make it bulletproof so when we see edge cases that might exist in the API that rust can't directly guard against we should strive to cover those with tests.

roslibrust/src/rosbridge/mod.rs
ssnover: Took a quick swing at this and it took more than a minute so I'll pick it up in another branch.

ssnover:
This is no longer work in progress and should be ready for a full review!

Carter12s:
Did you intend for this to Merge into master?

ssnover:
Oh yeah, I targeted at the other branch because that branch was still in review. It looks like we need to close branches when they're merged since we have a lot of stale branches.

ssnover:
The diff is much cleaner now actually. That's a little unexpected.

PR #38, opened 2022-09-10
Fix service file parsing

ssnover:
Made the searching for `---` more intelligent so that it won't fail if its in a comment. Add to AddTwoInts.srv so that we make sure this case can't regress.


ssnover:
Currently targeting `client-lifetime` but after that's merged can target `master`.

PR #40, opened 2022-09-11
Minor follow ups from recent MRs

Carter12s:


CHANGELOG.md
```diff
@@ -9,19 +9,21 @@ Steps:
...
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Current - Unreleased
 
 ### Changed
- - `Client` is now `ClientHandle`.
+ - `Client` is now `ClientHandle`
+ - All identifiers in generated code are now escaped with `r#`
```
ssnover, line 21: I would just note that this doesn't actually change anything as a user; i.e. no API break here.

ssnover (approved):
All looks good, but I just noted that adding `r#` isn't an API break of any kind, it won't break any user code.

CHANGELOG.md
Carter12s: Ahhh actually I think this is a really critical point.

I think a bit like rustfmt we should consider anything which affects the text of generated output code as a breaking API change.

Reason being: A unit test, exactly like the one we have, that re-run's code generation and checks the characters match would fail.

While it doesn't break the "Rust" API it does break the "source code" API. (I think this is part of why a lot of libraries don't support this style of generation).

If / when we make tutorials that show / recommend using build.rs and/or git to commit generated source code then anytime we change what the generated output would be we've broken our external API slightly. If we restrict ourselves to "macro-only" invocation then changing the resulting output isn't an API break.

With the non-macro generation style floating around we've also exposed ourselves to a slight edge-case which is: the crate being invoke on a message type which was generated from a previous version of the library. I still think it is worth supporting this style of invocation (and frankly recommending it) all of this is just trying to build the argument that things which affect the generated output should get documented as "Changed".

In the future I'm not sure I have answer as to how I would interpret a change like this wrt SemVer. Hopefully, we never have to do that because the generated code really shouldn't have to change much past this...

Rambles about what might cause it to change further:
* Support for actions? Pretty unlikely looks like rosbridge doesn't support it?
* Support for comments. Yeah we should do this at some point and move the comments from the .msg file into the generated output
* Support for MD5. Yeah we probably have to generate this at message gen time and stash on type, will add another trait impl for each message
* Does ROS2 need anything?

ssnover: I consider an API change to be something which you could detect due if you compiled a library and compared the symbols. An API break is something you can detect AND it breaks the build (i.e. the change was not purely additive). Frankly, our test is just to make sure the codegen is up to date so that it doesn't change when we make seemingly unrelated changes and happen to re-run codegen.

* Actions wouldn't break the API and could theoretically be supported transparently by writing an actionlib on our side. Underneath the ROS build generates msg files for each and at runtime an action client/server utilizes five subtopics (similar to how image_transport works).
* Comments are not an API break per above; doesn't go in the compiled binary.
* MD5 is additive, so we could roll that out in a minor release.
* No idea about ROS2.

PR #49, opened 2022-10-06
Moved over to using assets / submodules

### LINKED ISSUE #9: Recommendation: ros msg files as submodules instead of copies
Currently all of the ROS msg files that are frequently distributed with ROS are contained in the local_msgs directory of this repo. I think it'd make sense to include these via submodules, not because they are subject to change at some point, but just to direct back to a single source of truth.

Carter12s:
Closes #9


Carter12s:
Okay this ready for actual review. I will point out this change has now made our CI almost 1 minute slower (50% Slower). Maybe this will get faster one images are cached, but I think we increased our image size quite a bit and this time shows up in pulling the container.

PR #64, opened 2022-11-17
Huge re-org an a ton of fixes

Carter12s:


roslibrust/src/rosapi/mod.rs
```diff
@@ -6,8 +6,10 @@
 use crate::{ClientHandle, RosLibRustResult};
 use async_trait::async_trait;
 
-pub mod types;
-use types::*;
+// TODO Fix this... this sucks
+roslibrust_codegen_macro::find_and_generate_ros_messages_relative_to_manifest_dir!(
+    "../assets/ros1_common_interfaces/rosapi"
```
ssnover, line 11: Can you not use concat! and env!("CARGO_MANIFEST_DIR")?

ssnover: For reference: https://doc.rust-lang.org/cargo/reference/environment-variables.html

Carter12s: I cant!

At least it didn't seem to work when I tried

As far as I can tell the proc_macro is literally parsting the tokens passed into it as a vec of paths, and I can't pass in evaluatable code? am I crazzy?

ssnover: Oh yeah it looks like I did this with the function backed by this macro, not the macro itself.

PR #65, opened 2022-11-28
Fix char

Carter12s:
char != char

:frowning:


ssnover (approved):
Ah yeah, small oversight on my part.

PR #66, opened 2022-11-30
Initial Basic Support for ROS2 Message Generation

Carter12s:
Adds logic to parse package.xml and determine if a package is ROS1 or ROS2.
Fixes logic around determining if a line is a field and constant that was broken in ROS2 ( '=' could appear inside the type)
Begins to plumb through &Package to many of the parsing functions so they can decide on ROS1 vs. ROS2 behavior

This gets us to MPV status on handling ROS2 message files, the parser no longer panics while processing them. Lot's still to go.


README.md
```diff
@@ -10,28 +10,28 @@ Note on documentation:
...
-| rosapi                       | Planned                                                     |
+| tutorials                    | In Progress |
+| message_gen                  | ✅ |
+| advertise                    | ✅ |
+| unadvertise                  | ✅ |
+| publish                      | ✅ |
+| subscribe                    | ✅ |
+| unsubscribe                  | ✅ |
+| services                     | ✅ |
+| fragment / png / cbor        | ✅ |
```
ssnover, line 30: Did I miss when these became supported?

roslibrust_codegen/src/utils.rs
```diff
@@ -129,3 +149,58 @@ fn message_files_from_path(path: &Path, ext: &str) -> io::Result<Vec<PathBuf>> {
...
+        reader,
+        ParserConfig {
+            trim_whitespace: true,
+            ignore_comments: true,
+            ..Default::default()
+        },
+    );
+
+    let mut in_tag = false;
+    for e in parser {
```
ssnover, line 173: I remember doing this when trying to detect metapackages and I truly hate this API. It looks useful for streamed data, but it's such a pain to have to do state tracking manually.

roslibrust_codegen/src/utils.rs
```diff
@@ -129,3 +149,58 @@ fn message_files_from_path(path: &Path, ext: &str) -> io::Result<Vec<PathBuf>> {
...
+                }
+            }
+            Ok(XmlEvent::Characters(data)) => {
+                if in_tag {
+                    log::debug!("Got data inside of {TAG}: {data}");
+                    match data.as_str() {
+                        "catkin" => {
+                            return Ok(RosVersion::ROS1);
+                        }
+                        "ament_cmake" => {
```
ssnover, line 187: Is it also possible for this to be colcon or does this imply colcon? (not actually sure of the relationship between these tools)

ssnover (approved):

roslibrust_codegen/src/utils.rs
Carter12s: Yeah I'm defo not impressed, but it was pretty quick to get it to do what I wanted...

roslibrust_codegen/src/utils.rs
Carter12s: I asked this earlier today:
https://answers.ros.org/question/410017/how-to-determine-if-a-package-is-ros1-or-ros2/

We'll see if I can get and answer, I did the best digging I could through the ROS documentation and couldn't find a better answer. As far as I can tell this is never colcon, and instead colcon -> ament is the same relations ship as `catkin build` to `catkin_make`. I think colcon just invokes ament for you...

I'm not very sure about any of this hence the failure path in the code and the question...

PR #68, opened 2022-12-01
Working CI tests for ROS2

### LINKED ISSUE #42: CI for ROS2 / Full ROS2 support
I'm pretty dang sure this library is ROS2 compatible / compliant because rosbridge looks basically identical, but I'd like to confirm that.

Build a 2nd CI docker image that has rosbridge (also checkout #41 and probably stick rosapi node in as well), and make a second github action to run CI for both ros1 and ros2.

While doing this, think about how maybe to support different versions of ros2 and what generating that CI docker image would look like for Rolling vs. other versions. Is there anyway to auto update the image / kick-off CI as rolling is updated? I think a problem we're going to run into here is finding the right base images to build off of? If there is a "good" ROS2_rolling:latest image we can use this might not be terrible, but we likely need to strip 95% of the contents out of that image.

Bonus: update the auto-badges in the readme to show CI status for ROS1 and ROS2 (or even better various ROS2 versions).

Carter12s:
Closes #42

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -33,7 +33,7 @@ jobs:
       - name: Start rosbridge
         run: source /opt/ros/noetic/setup.bash; roslaunch rosbridge_server rosbridge_websocket.launch & disown; rosrun rosapi rosapi_node & sleep 1
       - name: Integration Tests
-        run: source /root/.cargo/env; cargo test --features running_bridge -- --test-threads 1
+        run: source /root/.cargo/env; cargo test --features ros1_test -- --test-threads 1
   galactic:
     runs-on: ubuntu-latest
     defaults:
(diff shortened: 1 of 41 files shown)


ssnover (approved):
I'm not particularly familiar with the ros2 launch process, but I'll trust the CI is actually doing something. What does `disown` do?

Carter12s:
> I'm not particularly familiar with the ros2 launch process, but I'll trust the CI is actually doing something. What does `disown` do?

I can confirm tests are running and failing if they actually don't work.

`disown` is releasing the process to run in the background so that it outlives the shell, it is just a bash command and nothing ros specific. I believe its the equivalent of "CTRL+Z". Not sure this is the best / cleanest bash way of achieving this, but with `disown` these processes actually continuing running in the background after this line and are available for the next command.

PR #75, opened 2023-01-25
Split codegen into two public functions with the option to call them separately

ssnover:
Per title.

I also refactored so that areas where a `BTreeMap<String, MessageFile>` was being returned will just return a `Vec<MessageFile>` now which means we don't have document the format of that `String`.

This will allow other possible code generation backends to be written; perhaps C++ and/or Typescript to throw two completely meaningless and randomly chosen examples out there.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
  - Support for default field values in ROS2 messages
+ - Added public APIs for getting message data from search and for generating Rust code given message data in roslibrust_codegen
 
 ### Fixed
  - Bug causing single quoted string constants in message files to not be parsed correctly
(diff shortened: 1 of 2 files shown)


ssnover:
By the way, I see we have some use of `log` inside this crate now. Do we want to switch all of the `eprintln` calls over to appropriate calls to that library? It might mean less helpful printed error messages for people using the macro, but we could always ask anyone posting an issue to run a non-macro version in order to get logs or to do initial debug.

ssnover:
By the way, prophetic comment on a flaky test I had to re-run:
```rust
// Impossibly short to actually work
let opts = ClientHandleOptions::new(LOCAL_WS).timeout(Duration::from_nanos(1));
assert!(ClientHandle::new_with_options(opts).await.is_err());
```

Carter12s:
> By the way, I see we have some use of `log` inside this crate now. Do we want to switch all of the `eprintln` calls over to appropriate calls to that library? It might mean less helpful printed error messages for people using the macro, but we could always ask anyone posting an issue to run a non-macro version in order to get logs or to do initial debug.

Yes! I think we should move over to log everywhere, I don't know about the proc_macro side of things to know enough about how to best provide debugging when things are invoked from there, but I assume log will be compatible with that.

PR #80, opened 2023-02-08
Refactor the pipeline for parsing ROS messages and services ahead of adding md5sum

ssnover:
Split our parsing up in order to represent the concept of a `ParsedMessageFile` vs a `MessageFile` such that the prior can be made before dependency resolution and the latter can eventually be made to hold more information that requires a resolved dependency graph.

I also made several clean ups around panics and split things into new functions. These could be tested individually if one was so inclined.

In my next PR, I'll be utilizing the new `MessageFile` struct to hold onto a few pieces of data including: `md5sum`, `has_header`, `is_fixed_size`, `is_message`, etc. which are things we'll need for ROS1 native and some incoming C++ code generation I'm working on.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
  - Support for default field values in ROS2 messages
  - Added public APIs for getting message data from search and for generating Rust code given message data in roslibrust_codegen
  - More useful logs available when running codegen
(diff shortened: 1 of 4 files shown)


Carter12s:
I'm not 100% sure quite what this achieves / what you are after in the end-game, but the changes are in-offensive and happy to have them. My gut says its dangerous to have partially resolved messages and that we want partially resolved messages to live for as short a time as possible and be as minimally exposed as possible... but I trust you have some idea of where you are headed.

Issue #73, opened 2023-01-19
the name `ApplyPlanningScene` is defined multiple times `ApplyPlanningScene` must be defined only once in the type namespace of this modulerustcClick for full compiler diagnostic


Akumar201:
Hi, I am using roslibrust to send data from a system that is having ROS install with moveit and gazebo. Following is the example I am following, however, I am getting the following error

```
the name `ApplyPlanningScene` is defined multiple times
`ApplyPlanningScene` must be defined only once in the type namespace of this modulerustc[Click for full compiler diagnostic](rust-analyzer-diagnostics-view:/diagnostic%20message%20%5B1%5D?1#file%3A%2F%2F%2Fhome%2Fakumar%2FTargetArm_Project%2Ftargetarm%2Fmessage_broadcaster%2Fsrc%2Fmain.rs)
the name `ChangeControlDimensions` is defined multiple times
`ChangeControlDimensions` must be defined only once in the type namespace of this modulerustc[Click for full compiler diagnostic](rust-analyzer-diagnostics-view:/diagnostic%20message%20%5B2%5D?2#file%3A%2F%2F%2Fhome%2Fakumar%2FTargetArm_Project%2Ftargetarm%2Fmessage_broadcaster%2Fsrc%2Fmain.rs)
the name `ChangeDriftDimensions` is defined multiple times
`ChangeDriftDimensions` must be defined only once in the type namespace of this modulerustc[Click for full compiler diagnostic](rust-analyzer-diagnostics-view:/diagnostic%20message%20%5B3%5D?3#file%3A%2F%2F%2Fhome%2Fakumar%2FTargetArm_Project%2Ftargetarm%2Fmessage_broadcaster%2Fsrc%2Fmain.rs)


```
```
use log::*;
use roslibrust::ClientHandle;

roslibrust_codegen_macro::find_and_generate_ros_messages!("/opt/ros/noetic/share/std_msgs");


#[tokio::main(flavor = "multi_thread")]
async fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
    simple_logger::SimpleLogger::new()
        .with_level(log::LevelFilter::Debug)
        .without_timestamps() // required for running wsl2
        .init()
        .unwrap();
```

    let client = ClientHandle::new("ws://localhost:9090").await?;
    let publisher = client.advertise::<std_msgs::String>("talker").await?;
    let val = "Hello World"; // sending example string to ros node working on another computer 
    loop {
[...]

Issue #54, opened 2022-11-03
 conflicting implementation for `realsense2_camera::DeviceInfo`


Akumar201:
Hi, I am trying to write a publisher to publish data to another computer where ROS will be running. However, I am getting the following error. I am sharing the code with you. I am getting the following error

```
error[E0428]: the name `DeviceInfo` is defined multiple times
 --> src/main.rs:6:1
  |
6 | roslibrust_codegen_macro::find_and_generate_ros_messages!();
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  | |
  | `DeviceInfo` redefined here
  | previous definition of the type `DeviceInfo` here
  |
  = note: `DeviceInfo` must be defined only once in the type namespace of this module
  = note: this error originates in the macro `roslibrust_codegen_macro::find_and_generate_ros_messages` (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0433]: failed to resolve: could not find `serde` in the list of imported crates
 --> src/main.rs:6:1
  |
6 | roslibrust_codegen_macro::find_and_generate_ros_messages!();
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ could not find `serde` in the list of imported crates
  |
  = note: this error originates in the macro `roslibrust_codegen_macro::find_and_generate_ros_messages` (in Nightly builds, run with -Z macro-backtrace for more info)

warning: unused import: `roslibrust_codegen_macro::find_and_generate_ros_messages`
 --> src/main.rs:3:5
  |
3 | use roslibrust_codegen_macro::find_and_generate_ros_messages;
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = note: `#[warn(unused_imports)]` on by default

error[E0119]: conflicting implementations of trait `roslibrust::RosServiceType` for type `realsense2_camera::DeviceInfo`
 --> src/main.rs:6:1
  |
6 | roslibrust_codegen_macro::find_and_generate_ros_messages!();
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  | |
  | first implementation here
[...]

PR #81, opened 2023-02-09
Generate md5sum

ssnover:
* Added calculation of md5sum. Checked a handful against generated header files, specifically `Image.msg` which includes an array and a Header and `TransformStamped.msg` which includes a Header and another msg datatype from the same package.
* Added helper methods throughout to make structs a little smarter and reduce duplicated code
* Adjusted pipeline to exclusively use ROS types and data directly derived from msg file during parsing and to not do any Rust code generation until we are doing generation (specifically around literals for constants).

### DIFF
--- a/roslibrust_codegen/src/parse.rs
+++ b/roslibrust_codegen/src/parse.rs
@@ -183,13 +183,13 @@ pub struct ParsedServiceFile {
 fn parse_field(line: &str, pkg: &Package, msg_name: &str) -> FieldInfo {
     let mut splitter = line.split_whitespace();
     let pkg_name = pkg.name.as_str();
-    let field_type = splitter.next().expect(&format!(
-        "Did not find field_type on line: {line} while parsing {pkg_name}/{msg_name}"
-    ));
+    let field_type = splitter.next().unwrap_or_else(|| {
+        panic!("Did not find field_type on line: {line} while parsing {pkg_name}/{msg_name}")
+    });
     let field_type = parse_type(field_type, pkg);
-    let field_name = splitter.next().expect(&format!(
-        "Did not find field_name on line: {line} while parsing {pkg_name}/{msg_name}"
-    ));
+    let field_name = splitter.next().unwrap_or_else(|| {
+        panic!("Did not find field_name on line: {line} while parsing {pkg_name}/{msg_name}")
+    });
 
     let sep = line.find(' ').unwrap();
     // Determine if there is a default value for this field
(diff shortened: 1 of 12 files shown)


roslibrust_codegen/src/parse.rs
```diff
@@ -331,29 +368,20 @@ fn strip_comments(line: &str) -> &str {
...
+        let pkg_version = pkg.version.unwrap_or(RosVersion::ROS1);
         FieldType {
-            package_name: if prop_map.contains_key(type_str) {
+            package_name: if is_intrinsic_type(pkg_version, type_str) {
                 // If it is a fundamental type, no package
-                if items[0] == "Header" {
+                None
+            } else {
+                // Otherwise it is referencing another message in the same package
+                if type_str == "Header" {
```
ssnover, line 380: Rather than treating Header as an internal type, I've changed over to explicitly assuming it comes from std_msgs here in all cases. This seems to be what some of the ROS source is doing.

ssnover:
@Carter12s All of the tests appear to be failing for what I think is the same reason: `test_msgs/Float64Stamped` has a dependency on `std_msgs/Header` which is no longer an intrinsic type (and shouldn't be). 

I could modify the pipeline for noetic to define `ROS_PACKAGE_PATH`, and then add ros1_test and ros2_test features similar to `roslibrust` tests. Is this your preferred approach?

ssnover:
This now fixes #82 #73 #54 all as an added bonus in my quest for passing CI.

roslibrust_codegen/src/parse.rs
```diff
@@ -26,7 +24,7 @@ lazy_static::lazy_static! {
         ("string", "::std::string::String"),
         ("time", "::roslibrust_codegen::integral_types::Time"),
         ("duration", "::roslibrust_codegen::integral_types::Duration"),
-        ("Header", "Header"),
+        //("Header", "Header"),
```
Carter12s, line 27: [Nit] delete?

roslibrust_codegen/src/utils.rs
```diff
@@ -145,6 +152,44 @@ fn message_files_from_path(path: &Path, ext: &str) -> io::Result<Vec<PathBuf>> {
     Ok(msg_files)
 }
 
+pub fn deduplicate_packages(packages: Vec<Package>) -> Vec<Package> {
```
Carter12s, line 155: Hrrmmm I'm slightly wondering if de-duplication at the package level is sane...

Like is it a valid use case to have

my_pkg/MessageA.msg and my_pkg/MessageB.msg and to actually want both to generate?

What you did is almost certainly fine, but has me wondering...

Carter12s:
Shouldn't this have update the generated code in roslibrust_test to include md5's for all our generated types? Why don't I see that change or a test failing because of an uncommited delta?

roslibrust_codegen/src/parse.rs
ssnover: From what I've seen, the ROS tools take the same approach.

roslibrust_codegen/src/utils.rs
ssnover: Where my_pkg has a different directory in each case with a package of the same name?

ssnover:
I added generation of the md5sum here, I did not add it to the types that are generated yet.

ssnover:
Something that's nasty is that we're starting to have a growth in the number of entry points for code generation which use different current working directories or which don't look at ROS_PACKAGE_PATH. It seems like it might be worth considering if we can reduce them down before releasing a 0.7.0.

PR #84, opened 2023-02-16
Add md5sum as associated constant to RosMessageType and RosServiceType

ssnover:
I have no idea if this valid or not for ROS2, but our code generation currently has no way to discern or behave differently (and when generating a module this means ROS1 and ROS2 types could easily be mixed...).

### DIFF
--- a/roslibrust_codegen/src/lib.rs
+++ b/roslibrust_codegen/src/lib.rs
@@ -128,10 +128,7 @@ pub struct ServiceFile {
 }
 
 impl ServiceFile {
-    pub fn resolve(
-        parsed: ParsedServiceFile,
-        graph: &BTreeMap<String, MessageFile>,
-    ) -> Option<Self> {
+    fn resolve(parsed: ParsedServiceFile, graph: &BTreeMap<String, MessageFile>) -> Option<Self> {
         if let (Some(request), Some(response)) = (
             MessageFile::resolve(parsed.request_type.clone(), &graph),
             MessageFile::resolve(parsed.response_type.clone(), &graph),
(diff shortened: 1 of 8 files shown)


roslibrust_test/src/ros1.rs
```diff
@@ -177,6 +151,47 @@ pub mod diagnostic_msgs {
...
+    #[allow(non_snake_case)]
+    #[derive(
+        :: serde :: Deserialize,
+        :: serde :: Serialize,
+        :: smart_default :: SmartDefault,
+        Debug,
+        Clone,
+        PartialEq,
+    )]
+    pub struct AddDiagnosticsRequest {
```
ssnover, line 165: Request and Response types will appear to be moved because I stopped adding them to the message graph (since a type can't depend on a service request or response type) and added them to the code generation of services.

roslibrust_codegen/src/lib.rs
```diff
@@ -98,21 +115,61 @@ impl MessageFile {
...
     pub(crate) parsed: ParsedServiceFile,
-    pub request: MessageFile,
-    pub response: MessageFile,
+    pub(crate) request: MessageFile,
+    pub(crate) response: MessageFile,
+    pub(crate) md5sum: String,
+}
+
+impl ServiceFile {
+    pub fn resolve(
```
ssnover, line 131: Whoops, need to make this private.

roslibrust_test/src/ros1.rs
```diff
@@ -28,6 +28,7 @@ pub mod actionlib_msgs {
     }
     impl ::roslibrust_codegen::RosMessageType for GoalID {
         const ROS_TYPE_NAME: &'static str = "actionlib_msgs/GoalID";
+        const MD5SUM: &'static str = "302881f31927c1df708a2dbab0e80ee8";
```
ssnover, line 31: Will add tests to the `roslibrust_test` crate in order to ensure that the ways these are calculated don't inadvertently change and that they match what's generated for the C++ headers.

PR #85, opened 2023-03-04
C++ service generation, adds constants to messages, and fixes some things with existing message generation, also CI

ssnover:
This MR adds:
* A jinja2 template for service headers
* Generation of headers for services,
* Generation of constant definitions
* Fixes to generation of message template, including translation of ros types to C++ types and recognition of ROS intrinsic types
* CI pipeline for verifying the message generation happened, was re-done in the event of changes, and that they will compile in a C++ project.


ssnover:
@Carter12s I'm going to need to some assistance bootstrapping the new container.

Carter12s:
I got the docker image generated and push. Found a few issues with the github CI action, but its still not passing. Seems like a diff is actually being detected, but the "Verify no changes" step gives you 0 debug info. Would encourage us to change that step to print the diff.

.github/workflows/noetic_gencpp.yml
```diff
@@ -0,0 +1,32 @@
+name: Noetic gencpp
```
Carter12s, line 1: Probably a TODO for later, but after getting more experience with github ci, seems like we should have just one workflow with each of these as seperate jobs in one file?

roslibrust_gencpp/README.md
```diff
@@ -0,0 +1,19 @@
...
+```bash
+roslibrust$ mkdir -p /tmp/sensor_msgs && \ 
+cargo run --bin gencpp -- \
+--msg assets/ros1_common_interfaces/common_msgs/sensor_msgs/msg/BatteryState.msg \
+--package sensor_msgs \
+-I std_msgs:assets/ros1_common_interfaces/std_msgs \
+-I geometry_msgs:assets/ros1_common_interfaces/common_msgs/geometry_msgs \
+-I sensor_msgs:assets/ros1_common_interfaces/common_msgs/sensor_msgs \
+--output /tmp/sensor_msgs
+```
```
Carter12s, line 14: Per my comment about the `git diff` line in CI, it might be really nice to have this CI be able to detect changes using the diffy library similar to how the other CI works.

roslibrust_gencpp/regenerate-test-msgs.sh
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env bash
```
Carter12s, line 1: I would prefer if this script was rust code instead of bash? Not a requirement, but I think not crazy?

Carter12s (approved):
Happy with all the code, CI stuff could still use some scrub, but like the changes

.github/workflows/noetic_gencpp.yml
ssnover: I'm not sure about best practices here either. Will look into after I get features merged.

roslibrust_gencpp/README.md
ssnover: Moved to cargo test.

roslibrust_gencpp/regenerate-test-msgs.sh
ssnover: I'm going to leave the script since it is easier to run and demonstrates the CLI, but the rust test will also generate the source to compare against.

PR #86, opened 2023-03-05
Support action files

ssnover:
This MR introduces parsing of action type messages into their seven generated messages and adding them to roslibrust codegen. Luckily, this doesn't require much additional work since they're just an abstraction over top of messages.

Based on the `genaction.py` script here: https://docs.ros.org/en/noetic/api/actionlib_msgs/html/genaction_8py_source.html


ssnover:
This should be reviewed after  #85 is merged.

PR #87, opened 2023-03-07
Initial cut of generic TopicProvider trait

Carter12s:


Carter12s:
Ended up getting to a good stopping point with pretty small changes. Figured baby steps on this one makes sense.

ssnover (approved):
It might be a good idea to add (or modify) an example that holds onto a `Box<dyn TopicProvider>` just to make sure it's sanely constructible.

PR #91, opened 2023-03-22
ROS1 Native Support

Carter12s:
Tracking MR for playing with ROS1 and seeing what it would take to have native support.


roslibrust/examples/native_ros1.rs
```diff
@@ -0,0 +1,88 @@
...
+    // println!("Result: {result:?}");
+
+    // rosrust uses a crate xml_rpc, but again no async
+    // Skipped testing it out
+
+    // 3rd try is the charm, serde_xmlrpc looks like it'll do the job, but we need to bring our
+    // own http client. Shouldn't be an issue just use hyper (nevermind use reqwest)
+    let client = roslibrust::MasterClient::new(
+        "http://localhost:11311",
+        "http://localhost:11312",
```
ssnover, line 25: From an API perspective it probably makes sense to use the new type pattern here since it will be easy to mix these up and isn't clear from looking at the invocation what each URI is for.

Carter12s: Yeah, frankly I'm still debating if I want this "raw client" to be a public API at all. Would love your thoughts on that.

I've been using this example as a cheap version of test driven development before I setup actual testing.

The ROS idiomatic action is for your node to automatically decide on its own what the second address is (the location your node is hosting its own server), but that really isn't very "Rust Like". Probably this all gets hidden under a client options builder pattern?

ssnover: I believe it just binds to port 0 and allows the kernel to decide, so I think it's definitely something that the user shouldn't have to know about. I'm not sure if it'll be necessary to parse the ROS_MASTER_URI variable in order to get the host or if we can just use localhost (probably fine for initial implementation).

roslibrust/src/ros1/master_client.rs
```diff
@@ -0,0 +1,360 @@
...
+        let response = self
+            .client
+            .post(&self.master_uri)
+            .body(request)
+            .send()
+            .await?
+            .text()
+            .await?;
+        trace!("Got response: {response}");
+        let parsed: (i8, String, T) = serde_xmlrpc::response_from_str(&response)?;
```
ssnover, line 76: Probably makes sense to destructure this tuple with a named variable for each? I don't know what these are by reading it here.

roslibrust/src/ros1/master_client.rs
```diff
@@ -0,0 +1,360 @@
...
+            "registerService",
+            vec![
+                self.id.clone().into(),
+                service.into().into(),
+                service_uri.into().into(),
+                self.client_uri.clone().into(),
+            ],
+        )?;
+        // Little conversion here to ignore third response parameter
+        let _: u8 = self.post(body).await?;
```
ssnover, line 117: Why specify the type for something that is being explicitly ignored?

roslibrust/src/ros1/master_client.rs
```diff
@@ -0,0 +1,360 @@
...
+        Ok(())
+    }
+
+    /// Hits the master's xmlrpc endpoint "unregisterService", returns true if the service was un-registered
+    /// and false if the master reports that the service was not registered as a provider
+    pub async fn unregister_service(
+        &self,
+        service: impl Into<String>,
+        service_uri: impl Into<String>,
+    ) -> Result<bool, RosMasterError> {
```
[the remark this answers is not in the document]
Carter12s: Yeah I debated this one... This actually exactly matches how the Server reports. Its a an API design question of whether attempting to unregister something which isn't already registered is actually an Error.

I decided to match the ROS api's style, and agree that "double un-registering" isn't actually a failure of the function. At the end of the day what the call represents "this service should not longer be called" is in effect, regardless of whether the service was possible to be called ahead of this.

roslibrust/src/ros1/master_client.rs
Carter12s: Actually have to for the templates to work out. ROS is returning a u8, but the api docs literally say that it should be ignored. If we don't parse the u8 the deserialization fails.

I think `self.post<u8>(body)` also works, but I kinda like this style?

roslibrust/examples/native_ros1.rs
Carter12s: Its actually worse than that:
http://wiki.ros.org/ROS/NetworkSetup

To "do it right" we actually have to look at two different optional Env vars: ROS_IP and/or ROS_HOSTNAME I'm a little unclear where specifically each should be used, but I think we should host our server at "ROS_IP:0" if it is set or at "ROS_HOSTNAME:0" if ROS_IP is not set.

ssnover:
For the sake of making review manageable I'd recommend making a feature branch for this as I'm sure it won't be trivial to keep CI happy as you develop which would make merging into the trunk onerous.

Carter12s:
Yeah good callout... I'll try to polish it to a mergeable chunk.

Do you prefer a feature branch or we can just merge to master and hide behind the ros1 feature flag? I'm actually probably okay with either.

ssnover:
If you're not having trouble keeping CI passing I think it's fine for it to go behind a feature flag on `master`.

roslibrust/examples/native_ros1.rs
```diff
@@ -0,0 +1,31 @@
+/**
+ * Testing ground for hitting on rosmaster directly
+ */
+
+#[cfg(feature = "ros1")]
```
ssnover, line 5: You can specify examples with required features: https://github.com/stm32-rs/stm32f1xx-hal/blob/master/Cargo.toml#L84

ssnover (approved):
There's too many changes for me to hold in my head at once. Since this is behind a feature flag, I'm okay with just merging for now and we'll shake bugs out as find them. I've reviewed the existing files and they look good.

PR #94, opened 2023-04-05
Generate with template

ssnover:
This MR reorganizes the genmsg crate in order to allow passing a custom template and a map of ros typenames to other typenames.


ssnover:
Hmm hold on this interface needs a lot of work...

Carter12s (approved):
Looks good!

PR #95, opened 2023-05-01
Add uint64 to ros type list and replace C++ mapping for string

ssnover:
* Updated the mapping for string type in C++ to exactly what ROS uses in order to allow overriding the allocator.
* Added forgotten `uint64` to list of ROS types.


Carter12s (approved):

Carter12s:
Curious to learn quite why ROS uses that string type and how overriding the allocator works...

ssnover:
`std::string` is actually just a convenient handle for `std::basic_string<char>`: https://en.cppreference.com/w/cpp/string/basic_string

PR #96, opened 2023-05-03
Add the ability to pass custom minijinja filters to the code generator

ssnover:
Figured out this API and it's pretty handy for some upcoming generation replacing some empy code with a ton of helpers.


Carter12s:
Looks good to me

PR #97, opened 2023-06-21
Print a polygon to stdout in order to break compilation

ssnover:


Carter12s:
The title of this MR leaves something to be desired...

roslibrust_genmsg/test_package/src/test_main.cpp
```diff
@@ -14,5 +14,7 @@ int main() {
     geometry_msgs::Polygon polygon;
     polygon.points.push_back(point);
 
+    std::cout << "Polygon: " << polygon << std::endl;
```
Carter12s, line 17: Just to confirm this is now an intentional part of the test to exercise the serialization portions?

Carter12s (approved):

roslibrust_genmsg/test_package/src/test_main.cpp
ssnover: Where the test is: does compilation fail or not. There's no verification of the output at all.

roslibrust_genmsg/test_package/include/sensor_msgs/BatteryState.h
```diff
@@ -342,89 +344,146 @@ namespace message_operations
...
+
     s << indent << "location: ";
-    s << std::endl;
     
     Printer< std::basic_string<char, std::char_traits<char>, typename std::allocator_traits<ContainerAllocator>::template rebind_alloc<char>>>::stream(s, indent + "  ", v.location);
     
+
+    
+    
+
```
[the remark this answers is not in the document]
ssnover: Yeah I'll take a swing at cleanup here. It can be a little hard to tune whitespace in these as minijinja is meant for generating HTML which most won't look at (and I think browsers auto-format if you do inspect it).

PR #99, opened 2023-06-28
Update parsing to support fixed-size arrays in ROS1

ssnover:
This MR introduces support for parsing fixed-size arrays and carrying their metadata around to the code generators. There's still a little support required in order to support fixed-size arrays in Rust code due to a lack of support in `serde`, however there is a possible workaround that I will investigate in a follow-up MR: https://github.com/serde-rs/serde/issues/631#issuecomment-322677033

The impetus for this was incorrect code generation for `sensor_msgs/CameraInfo` which has a relatively complex definition. As a result, I've added it to the `test_package` for C++. In order to detect regressions in checksum (and therefore parsing), I augmented some of the `ros::message_traits` structs/functions to use `constexpr` which allows checking the value in a `static_assert`. This should be backwards-compatible and purely an upgrade from a client code perspective.


Carter12s (approved):
LGTM

PR #103, opened 2023-08-13
Noetic compose now passes through master port for easier ros1 dev

Carter12s:


ssnover (commented):
Is this sufficient or will you need network_mode host in order to make it possible for it to contact your node's XML RPC server?

Carter12s:
This is sufficient for two nodes outside the docker image (on the host) to establish communication and talk with each other, brokering the communication through the master.

So this is enough to test roslibrust <-> roslibrust communication. If I wanted to talk with say rosapi inside the docker image, then this would not be enough.

I don't want to turn on network_mode: host for these files (its a copout). When we get a little farther I want to look into picking some specific ports for the nodes we care about and letting them through.

ssnover (approved):

PR #104, opened 2023-08-13
Improve test logging, fix a variety of compiler warnings

Carter12s:
Rolls out test_log usage to core roslibrust crate.
Fixes that we weren't testing with all features in CI (Major TODO to test matrix of features)
Fixes many warnings (but not all, because quite a few are legit)


ssnover (approved):
Looks good!

PR #105, opened 2023-08-13
Add definition to RosMessageType trait and generate Rust source to include it

ssnover:
Follow up to my PR from this morning which also adds the message definition to the Rust trait `RosMessageType` as this info is necessary in ROS1 native implementation.

### DIFF
--- a/roslibrust_codegen/src/gen.rs
+++ b/roslibrust_codegen/src/gen.rs
@@ -82,6 +82,7 @@ pub fn generate_struct(msg: MessageFile) -> TokenStream {
 
     let struct_name = format_ident!("{}", msg.parsed.name);
     let md5sum = msg.md5sum;
+    let definition = msg.parsed.source.trim();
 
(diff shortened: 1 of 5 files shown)


Carter12s:
So I'm going to approve, and I think you did the escaping correctly, but after seeing this fully rolled out I want to make a feature request. Fell free to ignore or throw on issues list.

I think if we switched from using "" to using the raw string literal syntax r#""# it would make the generated files look an order of magnitude cleaner as the DESCRIPTION field would actually be a readable string instead of a linebreak hell. It may also catch some strange string escaping edge cases, but I don't really think that is a valid justification.

Carter12s (approved):

ssnover:
I actually did no escaping here, it was done by the tokenization library. I'll spin it out into an issue. In general, people should probably look at the actual definition rather than the generated definition, but could be useful in some circumstances.

PR #109, opened 2023-08-27
ROS1 Native publishing first pass

ssnover:
Alright! So this has numerous issues; it will explode if you look at it wrong, but I successfully got the `ros1_talker` example to publish data to the listener node from the ROS1 tutorial: https://github.com/ros/ros_tutorials/blob/noetic-devel/rospy_tutorials/001_talker_listener/listener.py

Output from the listener node
```
demo_nodes-talker_listener-1  | [INFO] [1693161083.219396]: /listenerI heard hello world from rust 322
demo_nodes-talker_listener-1  | [INFO] [1693161084.220899]: /listenerI heard hello world from rust 323
demo_nodes-talker_listener-1  | [INFO] [1693161085.222399]: /listenerI heard hello world from rust 324
demo_nodes-talker_listener-1  | [INFO] [1693161086.227722]: /listenerI heard hello world from rust 325
demo_nodes-talker_listener-1  | [INFO] [1693161087.228721]: /listenerI heard hello world from rust 326
demo_nodes-talker_listener-1  | [INFO] [1693161088.231114]: /listenerI heard hello world from rust 327
demo_nodes-talker_listener-1  | [INFO] [1693161089.233616]: /listenerI heard hello world from rust 328
demo_nodes-talker_listener-1  | [INFO] [1693161090.235346]: /listenerI heard hello world from rust 329
```

Log output from the `ros1_talker` example
```
DEBUG [reqwest::connect] starting new connection: http://localhost:11311/
DEBUG [roslibrust::ros1::xmlrpc_server] New node xmlrpc connection AddrStream { inner: PollEvented { io: Some(TcpStream { addr: 127.0.1.1:39643, peer: 127.0.0.1:57142, fd: 12 }) }, remote_addr: 127.0.0.1:57142, local_addr: 127.0.1.1:39643 }
DEBUG [roslibrust::ros1::xmlrpc_server] requestTopic called by [String("/listener"), String("/chatter"), Array([Array([String("TCPROS")])])]
DEBUG [roslibrust::ros1::xmlrpc_server] Request for topic /chatter from /listener via protocols ["TCPROS"]
DEBUG [roslibrust::ros1::xmlrpc_server] Sending response for requested topic Response { status: 200, version: HTTP/1.1, headers: {}, body: Body(Full(b"<?xml version=\"1.0\" encoding=\"utf-8\"?><methodResponse><params><param><value><array><data><value><int>1</int></value><value><string>ready on pop-os:43577</string></value><value><array><data><value><string>TCPROS</string></value><value><string>pop-os</string></value><value><int>43577</int></value></data></array></value></data></array></value></param></params></methodResponse>")) }
INFO [roslibrust::ros1::publisher] Received connection from subscriber at 127.0.0.1:41932 for topic /chatter
DEBUG [roslibrust::ros1::publisher] Received subscribe request for /chatter
```

### DIFF
--- a/roslibrust/Cargo.toml
+++ b/roslibrust/Cargo.toml
@@ -63,7 +63,7 @@ running_bridge = []
 ros1_test = ["running_bridge"]
 # For use with integration tests, indicates we are testing integration with a ros2 bridge
 ros2_test = ["running_bridge"]
-# Provides access to experimental abstrac trait topic_provider
+# Provides access to experimental abstract trait topic_provider
 topic_provider = []
 # Provides a ros1 xmlrpc / TCPROS client
 ros1 = [
@@ -74,6 +74,11 @@ ros1 = [
     "dep:serde_rosmsg",
 ]
 
+
+[[test]]
+name = "ros1_xmlrpc"
+path = "tests/ros1_xmlrpc.rs"
+required-features = ["ros1_test", "ros1"]
 
 [package.metadata.docs.rs]
 features = ["all"]
--- a/roslibrust/Cargo.toml
+++ b/roslibrust/Cargo.toml
@@ -11,8 +11,10 @@ keywords = ["ROS", "robotics", "websocket", "json", "async"]
 categories = ["science::robotics"]
 
 [dependencies]
+abort-on-drop = "0.2"
 anyhow = "1.0"
 async-trait = "0.1"
+byteorder = "1.4"
 dashmap = "5.3"
 deadqueue = "0.2.4" # .4+ is required to fix bug with missing tokio dep
 futures = "0.3"
@@ -25,7 +27,12 @@ serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 smart-default = "0.6"
 thiserror = "1.0"
-tokio = { version = "1.20", features = ["net", "macros", "time", "rt-multi-thread"] }
+tokio = { version = "1.20", features = [
+    "net",
+    "macros",
+    "time",
+    "rt-multi-thread",
+] }
 tokio-tungstenite = { version = "0.17" }
 uuid = { version = "1.1", features = ["v4"] }
 roslibrust_codegen_macro = { path = "../roslibrust_codegen_macro", version = "0.7.0" }
@@ -33,7 +40,10 @@ roslibrust_codegen = { path = "../roslibrust_codegen", version = "0.7.0" }
 reqwest = { version = "0.11", optional = true } # Only used with native ros1
 # serde_xmlrpc = { version = "0.1.2", optional = true } # Only used with native ros1
 serde_xmlrpc = { git = "https://github.com/belak/serde-xmlrpc.git", branch = "main", optional = true } # Belak should be release v0.2 soon
-hyper = { version = "0.14", features = ["server"], optional = true } # Only used with native ros1
+serde_rosmsg = { version = "0.2", optional = true } # Only used with native ros1
+hyper = { version = "0.14", features = [
+    "server",
+], optional = true } # Only used with native ros1
 gethostname = { version = "0.4", optional = true } # Only used with native ros1
 
 [dev-dependencies]
@@ -56,7 +66,13 @@ ros2_test = ["running_bridge"]
 # Provides access to experimental abstrac trait topic_provider
 topic_provider = []
 # Provides a ros1 xmlrpc / TCPROS client
-ros1 = ["dep:serde_xmlrpc", "dep:reqwest", "dep:hyper", "dep:gethostname"]
+ros1 = [
+    "dep:serde_xmlrpc",
+    "dep:reqwest",
+    "dep:hyper",
+    "dep:gethostname",
+    "dep:serde_rosmsg",
+]
 
 
 [package.metadata.docs.rs]
--- a/roslibrust/examples/ros1_talker.rs
+++ b/roslibrust/examples/ros1_talker.rs
@@ -27,3 +27,8 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
 
     Ok(())
 }
+
+#[cfg(not(feature = "ros1"))]
+fn main() {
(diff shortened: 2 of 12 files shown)


ssnover:
Known problems:
1. I broke a bunch of tests which I want to rework into integration tests
2. Currently not possible to create multiple publishers for the same topic (previous will lose their connection on their mpsc channels)
3. Missing some connection header fields
4. Never calling `unregisterPublisher` on the rosmaster (not sure how to do this in a `Drop` implementation, maybe best effort `tokio::spawn`, move the rosmaster client in, and hope)

Carter12s:
DOPE!

Responses:
4. Probably best to set a flag or send a message via a channel on Drop and have an actor inside the client do the communication itself outside of the Drop path. Likely safer.
5. Yeah actually a good point, no idea how deep support for things like ~/chatter goes. I would have hoped that the rosmaster itself actually determined equivalence for things like that, but it probably does happen on the nodes side doesn't it.

Do you want to start merging this into a feature branch or do you plan to keep hacking here?

roslibrust/Cargo.toml
```diff
@@ -25,15 +27,23 @@ serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 smart-default = "0.6"
 thiserror = "1.0"
-tokio = { version = "1.20", features = ["net", "macros", "time", "rt-multi-thread"] }
```
Carter12s, line 28: Do you have a different toml linter than me? Does rustfmt hit the toml files? Just want to get us on same page.

roslibrust/examples/ros1_talker.rs
```diff
@@ -0,0 +1,32 @@
...
+        .unwrap();
+
+    let nh = NodeHandle::new("http://localhost:11311", "talker_rs")
+        .await
+        .map_err(|err| err)?;
+    let publisher = nh.advertise::<std_msgs::String>("/chatter", 1).await?;
+
+    let mut counter = 0u32;
+
+    while counter < 1000 {
```
Carter12s, line 21: really dumb nit, but I strongly prefer

`for counter in 0..1000`

roslibrust/src/ros1/node.rs
```diff
@@ -1,62 +1,361 @@
...
+            host_addr: addr,
+            hostname: hostname.to_owned(),
+            node_name: node_name.to_owned(),
+        };
+
+        tokio::spawn(async move {
+            loop {
+                match node.node_msg_rx.recv().await {
+                    Some(NodeMsg::Shutdown) => {
+                        log::info!("Shutdown requested, shutting down node");
```
Carter12s, line 220: Not sure if this log message was you or me, but we shouldn't log this if we're not actually going to shutdown node. This looks like a stub? Maybe shove in `unimplemented!`

roslibrust/src/ros1/publisher.rs
```diff
@@ -0,0 +1,166 @@
...
+                                let response_header_bytes = responding_conn_header
+                                    .to_bytes(false)
+                                    .expect("Couldn't serialize connection header");
+                                stream
+                                    .write(&response_header_bytes[..])
+                                    .await
+                                    .expect("Unable to respond on tcpstream");
+                                let mut wlock = subscriber_streams.write().await;
+                                wlock.push(stream);
+                            }
```
Carter12s, line 102: Defo need a log here in long run

roslibrust/src/ros1/tcpros.rs
```diff
@@ -0,0 +1,117 @@
...
+        let mut msg_definition = String::new();
+        let mut caller_id = String::new();
+        let mut latching = false;
+        let mut md5sum = String::new();
+        let mut topic = String::new();
+        let mut topic_type = String::new();
+        let mut tcp_nodelay = false;
+
+        // TODO: Unhandled: error, persistent
+
```
Carter12s, line 31: Prolly good to stick this link in here for future us:
http://wiki.ros.org/ROS/Connection%20Header

roslibrust/src/ros1/publisher.rs
```diff
@@ -0,0 +1,166 @@
...
+    pub(crate) fn new(topic_name: &str, sender: mpsc::Sender<Vec<u8>>) -> Self {
+        Self {
+            topic_name: topic_name.to_owned(),
+            sender,
+            phantom: PhantomData,
+        }
+    }
+
+    pub async fn publish(&self, data: &T) -> Result<(), Box<dyn std::error::Error>> {
+        let data = serde_rosmsg::to_vec(&data)?;
```
Carter12s, line 30: Holy shit... That just works. I expected there to be way more finagling there with some more complicated trait than `Serialize` being required.

Carter12s:
I left a bunch of comments, but all can be ignored if you want to merge. Also happy to start digging in on top of this if you want me to resolve a bunch of them?

roslibrust/Cargo.toml
ssnover: I will have to check what vscode extension I have, it very rarely changes the file, but pretty silly of us to keep changing it back and forth.

ssnover:
> I left a bunch of comments, but all can be ignored if you want to merge. Also happy to start digging in on top of this if you want me to resolve a bunch of them?

If you want to make cargo test ignore the failing 3 tests, I think this is sufficient to merge and hack on. I'm on a work trip all week so I won't have time to get back to this for a bit.

roslibrust/Cargo.toml
ssnover: I'm using Even Better TOML: https://marketplace.visualstudio.com/items?itemName=tamasfe.even-better-toml

roslibrust/src/ros1/node.rs
ssnover: The very next line is a break statement which will exit the loop and cause shutdown.

ssnover:
Carter I addressed your immediate feedback. For the points I've outlined, I address number 1. I'll make issues for each of the other ones after we merge this one.

PR #115, opened 2023-09-01
Enable multiple publishers on a single topic, guarded on same topic type str

### LINKED ISSUE #110: Enable creating multiple native ROS1 publishers for a single topic
Currently the only path to have multiple publishers for a topic is to clone the original handle. Even worse, attempting to make an additional one with `NodeHandle::advertise` will actually replace the old publisher in such a way that the old one is removed.

Goal here is for that to not happen.

ssnover:
Closes #110

### DIFF
--- a/roslibrust/src/ros1/node.rs
+++ b/roslibrust/src/ros1/node.rs
@@ -296,41 +296,14 @@ impl Node {
                 msg_definition,
                 md5sum,
             } => {
-                let channel = match PublishingChannel::new(
-                    &self.node_name,
-                    false,
-                    &topic,
-                    self.host_addr,
-                    queue_size,
-                    &msg_definition,
-                    &md5sum,
-                    &topic_type,
-                )
-                .await
+                match self
+                    .register_publisher(topic, &topic_type, queue_size, msg_definition, md5sum)
+                    .await
                 {
-                    Ok(channel) => channel,
-                    Err(err) => {
-                        let err_str = format!("Could not create publishing channel: {err:?}");
-                        log::error!("{err_str}");
-                        let _ = reply.send(Err(err_str));
-                        return;
-                    }
-                };
-                let pub_handle_sender = channel.get_sender();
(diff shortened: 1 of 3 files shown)


roslibrust_codegen_macro/src/lib.rs
```diff
@@ -46,9 +46,9 @@ pub fn find_and_generate_ros_messages_relative_to_manifest_dir(
 
     std::env::set_current_dir(env!("CARGO_MANIFEST_DIR")).expect("Failed to set working dir");
     for path in &mut paths {
-        *path = path
-            .canonicalize()
-            .expect(&format!("Failed to canonicalize path: {path:?}"));
+        *path = path.canonicalize().unwrap_or_else(|err| {
```
ssnover, line 49: This would create an error message if the resulting path didn't exist, but it didn't say why, now it does.

roslibrust/src/ros1/node.rs
```diff
@@ -372,6 +346,52 @@ impl Node {
...
+        queue_size: usize,
+        msg_definition: String,
+        md5sum: String,
+    ) -> Result<mpsc::Sender<Vec<u8>>, Box<dyn std::error::Error>> {
+        if let Some(handle) = self.publishers.iter().find_map(|entry| {
+            if entry.key().as_str() == &topic {
+                if entry.topic_type() == topic_type {
+                    Some(Ok(entry.get_sender()))
+                } else {
+                    Some(Err(Box::new(std::io::Error::from(
```
Carter12s, line 363: This error type isn't unreasonable, but I think it's very not human readable.

Fine to merge, but would prefer we give an error that clearly says "Attempted to register a publisher of a different type that was already registered to {topic}, new type, existing type"

Carter12s:
Looks good, happy to merge. Error types and messages need several overhauls, but good enough for now.

Carter12s (approved):

PR #116, opened 2023-09-02
Native subscribers are now supported!

ssnover:
Title.

### DIFF
--- a/roslibrust/src/ros1/node.rs
+++ b/roslibrust/src/ros1/node.rs
@@ -3,12 +3,10 @@
 
 use super::{
     names::Name,
-    publisher::{Publisher, PublishingChannel},
+    publisher::{Publication, Publisher},
     subscriber::{Subscriber, Subscription},
 };
-use crate::{
-    MasterClient, RosMasterError, ServiceCallback, Subscription, XmlRpcServer, XmlRpcServerHandle,
-};
+use crate::{MasterClient, RosMasterError, ServiceCallback, XmlRpcServer, XmlRpcServerHandle};
 use dashmap::DashMap;
 use roslibrust_codegen::RosMessageType;
 use std::net::{IpAddr, Ipv4Addr, ToSocketAddrs};
@@ -474,7 +472,7 @@ impl Node {
         }) {
             Ok(handle?)
         } else {
-            let channel = PublishingChannel::new(
+            let channel = Publication::new(
                 &self.node_name,
                 false,
                 &topic,
--- a/roslibrust/src/ros1/node.rs
+++ b/roslibrust/src/ros1/node.rs
@@ -1,14 +1,15 @@
 //! This module contains the top level Node and NodeHandle classes.
 //! These wrap the lower level management of a ROS Node connection into a higher level and thread safe API.
 
-use super::publisher::{Publisher, PublishingChannel};
-use crate::{
-    MasterClient, RosMasterError, ServiceCallback, Subscription, XmlRpcServer, XmlRpcServerHandle,
+use super::{
+    publisher::{Publication, Publisher},
+    subscriber::{Subscriber, Subscription},
 };
+use crate::{MasterClient, RosMasterError, ServiceCallback, XmlRpcServer, XmlRpcServerHandle};
 use dashmap::DashMap;
 use roslibrust_codegen::RosMessageType;
 use std::net::{IpAddr, Ipv4Addr, ToSocketAddrs};
-use tokio::sync::{mpsc, oneshot};
+use tokio::sync::{broadcast, mpsc, oneshot};
 
 #[derive(Debug)]
 pub struct ProtocolParams {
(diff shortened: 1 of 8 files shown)


roslibrust/src/ros1/node.rs
```diff
@@ -278,15 +312,19 @@ impl Node {
...
+                        if let Err(err) = subscription.add_publisher_source(&publisher_uri).await {
+                            log::error!(
+                                "Unable to create subscribe stream for topic {topic}: {err}"
+                            );
+                        }
                     }
-                };
-                item.known_publishers = publishers;
+                } else {
+                    log::warn!(
```
Carter12s, line 324: [nit] maybe log::debug!?

roslibrust/src/ros1/subscriber.rs
```diff
@@ -0,0 +1,219 @@
...
+use std::{marker::PhantomData, sync::Arc};
+use tokio::{
+    io::{AsyncReadExt, AsyncWriteExt},
+    net::TcpStream,
+    sync::{broadcast, RwLock},
+};
+
+pub struct Subscriber<T> {
+    receiver: broadcast::Receiver<Vec<u8>>,
+    phantom: PhantomData<T>,
```
Carter12s, line 13: [nit] _phantom?

roslibrust/src/ros1/subscriber.rs
```diff
@@ -0,0 +1,219 @@
...
+    pub(crate) fn new(receiver: broadcast::Receiver<Vec<u8>>) -> Self {
+        Self {
+            receiver,
+            phantom: PhantomData,
+        }
+    }
+
+    pub async fn next(&mut self) -> Result<T, Box<dyn std::error::Error>> {
+        let data = self.receiver.recv().await.map_err(|err| Box::new(err))?;
+        Ok(serde_rosmsg::from_slice(&data[..]).map_err(|err| Box::new(err))?)
```
Carter12s, line 26: So if I'm reading this right (which I may not be). If we have multiple subscribers to a topic, they each get a copy of the raw over the wire data and separately deserialize the data. That is totally fine for now, but would prefer if we deserde the data once and the held typed channels? I guess the complication there is that on the send side of the channel we need to know the type or create a type erasure?

Totally a valid impl for this stage, but maybe I'll file an issue to clean this up eventually.

roslibrust/src/ros1/subscriber.rs
```diff
@@ -0,0 +1,219 @@
...
+                    &node_name,
+                    &topic_name,
+                    &publisher_uri,
+                    connection_header,
+                )
+                .await
+                {
+                    publisher_list.write().await.push(publisher_uri.to_owned());
+                    // Repeatedly read from the stream until its dry
+                    let mut read_buffer = Vec::with_capacity(4 * 1024);
```
Carter12s, line 107: Something feels fishy here, and maybe I'm just not clear on the big picture.

It seems like a single call to read_buf could give a partial message, or multiple messages. From what I can tell whatever we shove down read_buf will attempted to be parsed as a single message, so i think there is a bug there?

I think we need to parse the ros header here, and packetize off of message length? Am I crazzy?

roslibrust/src/ros1/subscriber.rs
```diff
@@ -0,0 +1,219 @@
...
+    let bytes = stream.read_buf(&mut responded_header_bytes).await?;
+    if let Ok(responded_header) = ConnectionHeader::from_bytes(&responded_header_bytes[..bytes]) {
+        if conn_header.md5sum == responded_header.md5sum {
+            log::debug!(
+                "Established connection with publisher for {}",
+                conn_header.topic
+            );
+            Ok(stream)
+        } else {
+            log::error!("Tried to subscribe to {topic_name}, but md5sums do not match");
```
Carter12s, line 155: [nit] good to provide md5sum in log message

Carter12s (commented):
Would be willing to merge this as a solid start and keep hacking, but do have some concerns a step was missed.

roslibrust/src/ros1/node.rs
ssnover: I think this condition, while harmless, is indicative of a bug in our client so I'd prefer to keep the severity high.

roslibrust/src/ros1/subscriber.rs
ssnover: Right, the type erasure here is a little complicated and I decided to punt that problem since it's an optimization that's separate from setting up the subscriber.

roslibrust/src/ros1/subscriber.rs
ssnover: No, not crazy, that's totally possible. We're a little bit reliant on the other side of the stream sending valid data, but I do think we have to assume that and attempt to read the length to intelligently consume from the buffer. I'm not certain about the type and size of the buffer I'm using here. It's way overkill in a lot of cases, but it might add latency if a large single message got sent down the stream. On the other hand, I do think we want some kind of maximum in case a misbehaving node sends invalid data and we get tricked into trying to consume u32::MAX bytes or something.

ssnover: I'm inclined to say this works for small messages and we should fix this in a follow up MR.

PR #125, opened 2023-09-28
Working example of a generic subscriber

Carter12s:


ssnover (approved):
Novel use case, looks good.

Issue #122, opened 2023-09-27
Hung Integration Tests in Workflow


Carter12s:
Check here: https://github.com/Carter12s/roslibrust/actions/runs/6319225435/job/17195705451?pr=121

Job manually aborted after ~15 minutes. Needs a timeout in the test at a minimum and making the test fundamentally better as right result.

Carter12s:
Shane added the 20m limit here, but something is s still tripping it:
https://github.com/Carter12s/roslibrust/actions/runs/6360365907

Carter12s:
Last log points to:
![image](https://github.com/Carter12s/roslibrust/assets/8097919/df10cf0d-3698-4229-89de-cf20154cb6f5)

Carter12s:
Another example: 
![image](https://github.com/Carter12s/roslibrust/assets/8097919/b065b74a-3071-4cc2-9227-267cce222aeb)

https://github.com/Carter12s/roslibrust/actions/runs/6364358289/job/17280584950?pr=130

Carter12s:
Happened again post-merge so probably needs a deeper look 
https://github.com/Carter12s/roslibrust/actions/runs/6407918189/job/17395762347?pr=136

PR #132, opened 2023-09-30
Add timeouts to ros1 tests

Carter12s:
Closes: https://github.com/Carter12s/roslibrust/issues/122

Also nudged some other timeouts up. Just trying to keep CI jobs from hanging for 20 minutes.

Doesn't actually solve root problem likely, but makes failure quicker / more obvious (hopefully with better logs too).

### DIFF
--- a/roslibrust/src/rosbridge/integration_tests.rs
+++ b/roslibrust/src/rosbridge/integration_tests.rs
@@ -18,7 +18,7 @@ mod integration_tests {
     use tokio::time::{timeout, Duration};
     // On my laptop test was ~90% reliable at 10ms
     // Had 1 spurious github failure at 100
-    const TIMEOUT: Duration = Duration::from_millis(200);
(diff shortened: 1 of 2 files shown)


Carter12s:
Note: I suspect after these timeouts are added these new ROS1 tests are still flaky and we should probably add "may fail" to them for the time being.

ssnover (approved):
This looks good for now. I think we should consider defining a `roslibrust::test_with_timeout` macro so that the timeout doesn't need to be added everywhere, it's kind of noisy.

Carter12s:
> This looks good for now. I think we should consider defining a `roslibrust::test_with_timeout` macro so that the timeout doesn't need to be added everywhere, it's kind of noisy.

Yeah I ran into this same thing with the roslibrust client and ended up building it as a feature of the crate with timeout being configurable on the NodeHandle. Likely will want to do the same thing to the ros1 client eventually.

Issue #129, opened 2023-09-29
Failed to parse message file


Carter12s:
When testing today on calculon I tripped the following tests:

```
test test::generate_ok_on_ros1 ... FAILED
test test::generate_ok_on_ros2 ... FAILED

failures:

---- test::generate_ok_on_ros1 stdout ----
thread 'test::generate_ok_on_ros1' panicked at 'Failed to parse a literal value in a message file to the corresponding rust type: /bond_disable_heartbeat_timeout to String', roslibrust_codegen/src/gen.rs:262:85
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- test::generate_ok_on_ros2 stdout ----
thread 'test::generate_ok_on_ros2' panicked at 'Failed to parse a literal value in a message file to the corresponding rust type: /bond_disable_heartbeat_timeout to String', roslibrust_codegen/src/gen.rs:262:85

```

Appears to be caused by this message: http://docs.ros.org/en/diamondback/api/bond/html/msg/Constants.html

From this package:
https://wiki.ros.org/bond

Not sure how it made it on my system, but we should add it as an adversarial test case and support it.

Carter12s:
God the documentation on how this "should work" is really bad:
![image](https://github.com/Carter12s/roslibrust/assets/8097919/0d69cabe-59d3-40b2-8992-c6b5c2f134b6)

Carter12s:
Oh and ROS1 and ROS2 disagree with each other:
![image](https://github.com/Carter12s/roslibrust/assets/8097919/99735144-42dd-4955-8888-486823fb8c9e)

PR #127, opened 2023-09-28
Better errors from codegen

Carter12s:
Closes: https://github.com/Carter12s/roslibrust/issues/126
Closes: https://github.com/Carter12s/roslibrust/issues/128

Trying to catch some common user errors and make them more clear. Investigated automated tests for these kinds of errors, but it looks like it is actually pretty annoying to setup. Looks like this will do the job, but a little much to setup right now: https://crates.io/crates/compiletest_rs

These test checks were manually performed by me while setting this up:
Manually modifying calling_services to introduce a typo on this line:
`roslibrust_codegen_macro::find_and_generate_ros_messages!("assets/ros1_common_interfacs",);`

Now produces the following compilation error when invoking `cargo run --example calling_service`
```
error: Failed to find any services or messages while generating ROS message definitions, paths searched: ["assets/ros1_common_interfacs"]
 --> roslibrust/examples/calling_service.rs:4:1
  |
4 | roslibrust_codegen_macro::find_and_generate_ros_messages!("assets/ros1_common_interfacs",);
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = note: this error originates in the macro `roslibrust_codegen_macro::find_and_generate_ros_messages` (in Nightly builds, run with -Z macro-backtrace for more info)
```

PR #133, opened 2023-09-30
Fix handling of ROS1 string constants, add test case

Carter12s:
Closes: #129


Carter12s:
I still don't think this logic is perfect.

I think to technically be in compliance with ROS1 we would have to very intentionally not strip comments from lines that are string constants, but that is an edge case I just don't want to deal with today.

Carter12s:
Again, should be noted CI failure is not related to this changes, and addressed elsewhere.

roslibrust_codegen/src/gen.rs
```diff
@@ -257,10 +268,19 @@ fn parse_ros_value(ros_type: &str, value: &str, array_info: Option<Option<usize>
...
-                let parsed: String = serde_json::from_str(value).unwrap_or_else(|_| panic!("Failed to parse a literal value in a message file to the corresponding rust type: {value} to String"));
-                quote! { #parsed }
+                match version {
+                    RosVersion::ROS1 => {
+                        // For ROS1 then entire contents except for leading and trailing whitespace are used
+                        let value = value.trim();
+                        quote! { #value }
+                    }
+                    RosVersion::ROS2 => {
+                        // For ROS2 value must be in quotes, and either single or double quotes are okay
```
ssnover, line 278: Can a string be passed like `'this is a "string"'`

ssnover (approved):
Just asking about one more corner case in the string parsing that occurred to me during review.

roslibrust_codegen/src/gen.rs
Carter12s: Yeah we're almost certainly handling this wrong, but the ROS2 documentation is so vague about what correct behavior is I don't even know what to do:
https://docs.ros.org/en/rolling/Concepts/Basic/About-Interfaces.html?highlight=interface%20files#constants

image

The documentation literally doesn't say anything about string values for ROS2 being in quotes, and we have to assume that I guess from the default values section, but it says default values aren't escaped, but nothing about constant values. Again I guess we assume they are not. I think I'll take another swing at some more robust logic and probably add a set of unit tests around it, just to document what we think valid is.

I think the only way to figure out what is "actually correct here" is to look at what the ros2 tools do, but that seems like work...

ssnover: I'm not sure what the appropriate channel is to ask for clarification, but our best path would be to get that section of the documentation clarified (even if the result is that lack of definition means it could be anything).

Carter12s:
@ssnover I think the algorithm is now correct, but I want #127 to merge first so I can do proper error handling in this MR.

Will also probably add units test then too, annoying while it doesn't return result

PR #134, opened 2023-09-30
Update changelog

Carter12s:
I opened a spat of PRs today, I think they add enough value we should cut 0.8 so did some prep for that.

Really should have been updating changelog along the way.


ssnover (approved):
I'll add a PR template so that it's harder to forget to update the changelog.

Issue #120, opened 2023-09-26
Improve build.rs ergonomics


Carter12s:
Setting up a new project using this library I came up with this minimal build.rs file that allowed me to debug the crate:

```
fn main() {
    let _ = simple_logging::log_to_file("test.log", log::LevelFilter::Debug);

    let p = "<REDACTED>";
    cargo_emit::rerun_if_changed!(p);

    let tokens = roslibrust_codegen::find_and_generate_ros_messages(
        vec![p.into()],
    );

    let _ = std::fs::write("./src/msg.rs", tokens.to_string());
}
```

Several drawbacks / places for improvement here:

1. Would be really cool if we handled the cargo_emit stuff for the user (potentially behind a flag).
2. Would be cool if we handled the fs::write stuff for the user.
3. Really want to figure out how to make log output show up from either the proc macro or build.rs style in a more easy way. Rust compiler toolchain makes the hard (for good reason), but want to give users an easy way to flip a flag and get logging. Ideally we would create good error handling throughout the entire code gen process and bubble up meaningful errors, but short of that logging is the best thing we can do.
4. I think we need a tutorial / example setup of this documented somewhere.
5. The resulting code generated by this is not correctly formatted... Would love us to provide a reasonable solution here.
6. The generated code failed to compile because of the existing "leaked dep" issue, and needing to add serde, roslibrust_codegen, and smart-default as deps.

Need to pave road here more.

PR #135, opened 2023-09-30
Add an example for build.rs usage

Carter12s:
Closes: https://github.com/Carter12s/roslibrust/issues/120

Would not call this example "good" at this time.

Three big TODOs left:

1. Needs the list of directories for cargo to watch to be fed back by codegen stuff
2. Need to incorporate the error handling from the other MR
3. Would love to fix the leaky dependencies (not sure if we actually can in this case?)

Probably good enough to merge as is and fix those with later issues.


example_package/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
+name = "example_package"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+# This is what we need for using this package in our mono-repo
+roslibrust = { path = "../roslibrust" }
```
ssnover, line 8: You can pass both. Cargo will prefer paths if they're available over the version. Not sure if it's worth updating and adding a note about that here or not.

ssnover (approved):

example_package/Cargo.toml
Carter12s: Yeah will punt, I actually had a line with both in there, but figured that was actually more confusing as an example

Issue #131, opened 2023-09-30
Recent rust version may have effed with some relative paths


Carter12s:
I recently updated to 1.72 on one of the computers I develop on.

Compilation now fails with failure to find message files from some examples, tests, and doc comments. The failures all seem to originate from calls to the proc macro that use "../assets/" instead of "assets/" as the path to find the files. I suspect the "working directory" for test execution has changed?

Carter12s:
Documentation indicates that tests should always execute relative to package (not workspace) root:
https://github.com/rust-lang/cargo/issues/11852

But that doesn't seem the case at the moment on my system?

Carter12s:
Confirmed that compilation behavior is different between rust 1.71 and rust 1.72

Carter12s:
Looks like this might be the culprit:
![image](https://github.com/Carter12s/roslibrust/assets/8097919/7c350e3f-1900-4215-85bd-9ec5f9ba4562)

Carter12s:
Yeah looks like updated docs here:
https://doc.rust-lang.org/nightly/cargo/commands/cargo-test.html#working-directory-of-tests

Carter12s:
The bullet point in the release notes is actually a bit of a lie, because now the behavior appears to differ between actual tests and doc tests.

Actual tests appear to run from "Package Root" while doctests now appear to run from "Workspace Root"

We could fix this by either adding some commands to our tests invocation, or by just updating everything to rust 1.72 and standardizing? Probably prefer the later

PR #136, opened 2023-10-02
Fix build on 1 72

Carter12s:
Closes: #131

### DIFF
--- a/roslibrust/src/rosbridge/client.rs
+++ b/roslibrust/src/rosbridge/client.rs
@@ -53,7 +53,7 @@ impl ClientHandleOptions {
 /// ClientHandle is clone and multiple handles can be clone()'d from the original and passed throughout your application.
 /// ```no_run
 /// # roslibrust_codegen_macro::find_and_generate_ros_messages!(
-/// #    "../assets/ros1_common_interfaces/std_msgs"
+/// #    "assets/ros1_common_interfaces/std_msgs"
 /// # );
 /// # #[tokio::main]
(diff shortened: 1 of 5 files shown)


Carter12s:
So CI passing confirms that this does fix it, but I'm a little bit torn on if this is the actual solution I want. This means doc-tests can only be built on >1.72 which is messy. We could probably come up with a solution that works both pre and post, but not sure it is actually worth doing? @ssnover thoughts?

roslibrust/src/rosbridge/client.rs
```diff
@@ -53,7 +53,7 @@ impl ClientHandleOptions {
 /// ClientHandle is clone and multiple handles can be clone()'d from the original and passed throughout your application.
 /// ```no_run
 /// # roslibrust_codegen_macro::find_and_generate_ros_messages!(
-/// #    "../assets/ros1_common_interfaces/std_msgs"
```
ssnover, line 56: Can we `concat!` the workspace path and this path so that it's always a relative path? That should work for all versions.

ssnover (commented):
I think we just need to use an absolute path if we care to support pre-1.72.

roslibrust/src/rosbridge/client.rs
Carter12s: Huh not actually sure if we can do that with workspace path. No workspace path var listed here: 
https://doc.rust-lang.org/cargo/reference/environment-variables.html
Looks like issue to add that variable here:
https://github.com/rust-lang/cargo/issues/3946

I guess we could concat off of CARGO_MANIFEST_DIR and that would work in both versions. Probably better that way? I'll try that out

Issue #118, opened 2023-09-21
Overhaul errors for native ROS1 implementation


ssnover:
Lots of Boxed dyn Errors, it's a mess.

PR #140, opened 2023-10-05
Trying to figure out which tests are flaky and fix, adding debug info…

Carter12s:
## Description
Trying to make CI more robust

## Fixes
Closes: #122
Closes: #141

## Checklist
- [ ] Update CHANGELOG.md


Carter12s:
https://github.com/Carter12s/roslibrust/actions/runs/6451070088/job/17511214836?pr=140

image

Carter12s:
https://github.com/Carter12s/roslibrust/actions/runs/6451070091/job/17511214844?pr=140
image

Carter12s:
Okay @ssnover I feel like I'm just summoning ghosts at this point.

If you have a chance to review the most recent test failure here: https://github.com/Carter12s/roslibrust/actions/runs/6459122424/job/17540392614?pr=140

Examine the current design of that test on https://github.com/Carter12s/roslibrust/pull/140/commits/4c858f01d611fdab9a1c5dc2c5a545b322e1271e

And help me come up with theories on HOW THE FUCK this test could still be timing out without printing anything.

Here is my current mental model:
 - The log line `test tests::verify_get_publications has been running for over 60 seconds` indicates to me that this test must be the problem, and that the overall test runner is healthy.
 - Every single `async` operation within that test is wrapped in a timeout()
 - The overall test has a watchdog configured on it. I've manually confirmed the watchdog will trip anything in the body of the test is delayed
 - The test appears to run perfectly on my system and only fail in CI (ran it in a loop in bash, 100/100 times passed). Althou I wasn't re-building between runs.
 - The test only sometimes fails in CI
 - tokio::test is defaulting to a single threaded running. I'll try multi next, but that would mean that somehow some function we're calling within the test is blocking the only tokio thread preventing the watchdog?

Carter12s:
Found this issue: https://github.com/ZcashFoundation/zebra/issues/1631

.to_socket_addrs() could be the blocking culprit.

Carter12s:
JESUS CHRIST HOW DOES IT STILL TIME OUT

Carter12s:
Leaving a note here before I forget:

We are spawning a tokio task which is never dying, and isn't being killed on drop, panics! aren't propagating to it, and it is keeping the test process alive (We think this is the root cause).

Carter12s:
Okay I think I've finally finally actually traced down the root problem, had to fall back to "log debugging" cause I could neither get a debugger to produce reliably, or get all my mucking about with timeouts to work (still unclear why timeouts aren't being effective), but whatever there is a root bug.

Inside of node.rs/registerPublisher:
```rust
            log::trace!("Created new publication for {topic:?}");
            let handle = channel.get_sender();
            self.publishers.insert(topic.clone(), channel);
            log::trace!("Inserted new publsiher into dashmap");
```

When the test verify_get_publications hangs it reliably publishes the first log message and not the second. The DashMap is what is causing a deadlock, and it makes sense now that it is probabilistically occurring since the deadlock will only occur (I believe) if two objects end up in the same DashMap shard...

Will continue debugging and try to solve root issue. Unclear what I'm ultimately going to do with this branch as it has turned in a giant set of changes as I randomly poked at things to find the root issue.

Carter12s:
Okay @ssnover I think this is finally ready to merge. Dashmap was the root of the problem as far as I can tell.

CHANGELOG.md
```diff
@@ -30,6 +30,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+ - Removed `find_and_generate_ros_messages_relative_to_manifest_dir!` this proc_macro was changing the current working directory of the compilation job resulting in a variety of strange compilation behaviors. Build.rs scripts are recommended for use cases requiring fine
```
ssnover, line 33: I'm not sure it was necessary to remove this macro entirely. It just didn't need to be changing the working directory. Like it just need to prepend the provided paths in the macro with the manifest directory path.

roslibrust/examples/native_ros1.rs
```diff
@@ -4,7 +4,7 @@
 
 #[cfg(feature = "ros1")]
 #[tokio::main]
-async fn main() -> Result<(), Box<dyn std::error::Error>> {
+async fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
```
ssnover, line 7: Why make the return of `main` have these bounds? I think it's just noise.

roslibrust/src/ros1/publisher.rs
```diff
@@ -26,8 +28,10 @@ impl<T: RosMessageType> Publisher<T> {
         }
     }
 
-    pub async fn publish(&self, data: &T) -> Result<(), Box<dyn std::error::Error>> {
-        let data = serde_rosmsg::to_vec(&data)?;
+    pub async fn publish(&self, data: &T) -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
```
ssnover, line 31: If a RosLibRustError is being mapped here, just use that type instead of boxing it into something with dyn trait bounds.

roslibrust/src/rosbridge/integration_tests.rs
```diff
@@ -135,9 +135,13 @@ mod integration_tests {
         // Intentionally a port where there won't be a server at
         let opts = ClientHandleOptions::new("ws://localhost:9091").timeout(TIMEOUT);
         assert!(ClientHandle::new_with_options(opts).await.is_err());
+
+        // This case sometimes actually passed in CI and caused test to fail
```
ssnover, line 139: I think if we want to keep this case it could be better served by directly spawning with `tokio::task::spawn` but not awaiting it, yielding back to tokio manually (`tokio::task::yield_now()`), then manually calling poll on the handle returned from spawning. I think this allows forcing the timeout.

roslibrust/tests/ros1_xmlrpc.rs
```diff
@@ -52,23 +44,19 @@ mod tests {
...
-    #[tokio::test]
-    async fn verify_get_publications() {
-        let node = timeout(
-            TIMEOUT,
-            roslibrust::NodeHandle::new("http://localhost:11311", "verify_get_publications"),
-        )
-        .await
-        .unwrap()
-        .unwrap();
+    #[test_log::test(tokio::test(flavor = "multi_thread"))]
```
ssnover, line 52: Does this still require the multi thread flavor and the test_log macro?

roslibrust/tests/ros1_xmlrpc.rs
```diff
@@ -94,26 +78,27 @@ mod tests {
             vec!["/verify_get_publications".into()],
         )
         .await;
+        log::info!("Got post advertise publications");
+
         assert_eq!(publications.len(), 1);
-        let (topic, topic_type) = publications.iter().nth(0).unwrap();
+        let (topic, topic_type) = publications
```
ssnover, line 84: Is this modification still necessary?

ssnover (commented):
A lot of cruft built up while debugging this test I think.

ssnover:
I don't see any changes related to `DashMap`, can you explain?

roslibrust/src/rosbridge/integration_tests.rs
Carter12s: Agreed, but yeah I don't think we need it re-enabled at the moment.

roslibrust/tests/ros1_xmlrpc.rs
Carter12s: 1. I think all of our tests should be wrapped in test_log

2. Maybe, have tried a lot of permutations of testing, but it is hard to get the failures to manifest... I don't think it is harmful being in there for now.

roslibrust/tests/ros1_xmlrpc.rs
Carter12s: Not sure... Potentially.

Generally I found some pretty horrible behavior with any kind of panic! not cleaning things up correctly. I'm not sure if the root problem was the DashMap and now we have good panic behavior or not.

roslibrust/src/ros1/publisher.rs
Carter12s: Punt to #118

roslibrust/examples/native_ros1.rs
Carter12s: I think we should punt to #118 

I ended up having to add this bound in a number of places to make it valid to return the errors from tasks.

CHANGELOG.md
Carter12s: I want to remove because I do think it is just a bad API (like the method name is a joke) we could have fixed it, but I really think this macro came about from me not understanding aspects of when macros were expanded and what working directory / cargo manifest dir were under different circumstances.

I think ultimately we figure out a new API in the future for the same reason.

roslibrust/src/ros1/node.rs
```diff
@@ -229,11 +243,11 @@ pub struct Node {
     // Receiver for requests to the Node actor
     node_msg_rx: mpsc::UnboundedReceiver<NodeMsg>,
     // Map of topic names to the publishing channels associated with the topic
-    publishers: DashMap<String, Publication>,
+    publishers: HashMap<String, Publication>,
```
Carter12s, line 246: @ssnover DashMap changes are here.

Carter12s:
> I don't see any changes related to `DashMap`, can you explain?

I think you missed reviewing node.rs (needed to expand the diff). Essentially I hand originally put DashMaps as the core data storage mechanism in the ros1::node to allow concurrent access. However, you changed the execution pattern so that node is only ever being accessed by the single "node task" that is receiving from the socket, so the concurrency protections of the DashMap isn't needed (no one actually needs simultaneous mutable access to Node).

Somehow, (and I really am not 100% sure how) DashMap and tokio were fighting with each other, and the DashMap was ending up in a deadlock, even thou its only ever being accessed by one "task". The DashMap deadlocking was (I believe) the root problem to the timeouts / watchdogs not working because Drop for Node was deadlocking as the DashMap couldn't be destructed...

Carter12s:
> A lot of cruft built up while debugging this test I think.

You don't even know man... This is after I cleaned up quite a bit...

ssnover:
Makes sense to remove the DashMap in that context, though I'm not convinced we know the root cause fully.

If there isn't a pressing need to put out a new release, can you use test_log throughout and remove some of the changes that were sanity checks? I think the error stuff can be punted, but I don't think we need to add new cruft unnecessarily.

ssnover (approved):
Thanks for humoring me! Nice job tracking down the issues!

Issue #148, opened 2024-01-28
ROS1 MAVROS Decode Failure: Decoded data is shorter than predicted value length


your-sudden-death:
Hi everyone,
when trying to decode a `/mavros/global_position/global` (type `NavSatFix`) message with this crate (using ROS1 natively),
the decoder fails with
```
2024-01-28T14:23:11.616115Z DEBUG reqwest::connect: starting new connection: http://localhost:11311/
2024-01-28T14:23:11.616335Z DEBUG hyper::client::connect::dns: resolving host="localhost"
2024-01-28T14:23:11.616598Z DEBUG hyper::client::connect::http: connecting to 127.0.0.1:11311
2024-01-28T14:23:11.616710Z DEBUG hyper::client::connect::http: connected to 127.0.0.1:11311
2024-01-28T14:23:11.616991Z DEBUG hyper::proto::h1::io: flushed 254 bytes
2024-01-28T14:23:11.617371Z DEBUG hyper::proto::h1::io: parsed 4 headers
2024-01-28T14:23:11.617387Z DEBUG hyper::proto::h1::conn: incoming body is content-length (258 bytes)
2024-01-28T14:23:11.617418Z DEBUG hyper::proto::h1::conn: incoming body completed
2024-01-28T14:23:11.617477Z DEBUG hyper::client::pool: pooling idle connection for ("http", localhost:11311)
2024-01-28T14:23:11.617823Z DEBUG hyper::client::pool: reuse idle connection for ("http", localhost:11311)
2024-01-28T14:23:11.617908Z DEBUG hyper::proto::h1::io: flushed 466 bytes
2024-01-28T14:23:11.618458Z DEBUG hyper::proto::h1::io: parsed 4 headers
2024-01-28T14:23:11.618465Z DEBUG hyper::proto::h1::conn: incoming body is content-length (277 bytes)
2024-01-28T14:23:11.618481Z DEBUG hyper::proto::h1::conn: incoming body completed
2024-01-28T14:23:11.618515Z DEBUG hyper::client::pool: pooling idle connection for ("http", localhost:11311)
2024-01-28T14:23:11.618731Z DEBUG hyper::client::pool: reuse idle connection for ("http", localhost:11311)
2024-01-28T14:23:11.618792Z DEBUG hyper::proto::h1::io: flushed 483 bytes
2024-01-28T14:23:11.619146Z DEBUG hyper::proto::h1::io: parsed 4 headers
[...]

PR #150, opened 2024-02-08
Support for Fixed Sized Arrays

Carter12s:
## Description
ROS1 serialization with serde_rosmsg requires that fixed sized arrays be represented not as Vec<_> but as [_; N]. The two are sent over the wire differently. Most of the support for this was already plumbed through, so this just closed the loop at the end, but is hitting the "const generic" barrier.

## Fixes
Closes: #148 

## Checklist
- [x] Update CHANGELOG.md

### DIFF
--- a/roslibrust/Cargo.toml
+++ b/roslibrust/Cargo.toml
@@ -45,6 +45,8 @@ hyper = { version = "0.14", features = [
 ], optional = true } # Only used with native ros1
 gethostname = { version = "0.4", optional = true } # Only used with native ros1
 regex = { version = "1.9", optional = true } # Only used with native ros1
+# TODO I think we should move rosapi into its own crate...
+serde-big-array = { version = "0.5", optional = true } # Only used with rosapi
 
 [dev-dependencies]
 env_logger = "0.10"
@@ -57,7 +59,7 @@ default = []
 # Note: all does not include running_bridge as that is only intended for CI
 all = []
 # Provides a rosapi rust interface
-rosapi = []
+rosapi = ["serde-big-array"]
 # Intended for use with tests, includes tests that rely on a locally running rosbridge
 running_bridge = []
 # For use with integration tests, indicating we are testing integration with a ros1 bridge
--- a/roslibrust/Cargo.toml
+++ b/roslibrust/Cargo.toml
@@ -50,6 +50,7 @@ regex = { version = "1.9", optional = true } # Only used with native ros1
 env_logger = "0.10"
 test-log = "0.2"
 simple_logger = "2.1.0"
+serde-big-array = "0.5"
 
 [features]
 default = []
--- a/roslibrust_codegen/src/gen.rs
+++ b/roslibrust_codegen/src/gen.rs
@@ -173,16 +173,15 @@ fn generate_field_definition(
             quote! {}
         }
     };
-    let serde_line = if let Some(Some(fixed_array_length)) = field.field_type.array_info {
-        if fixed_array_length > 32 {
+    // This is the largest size of fixed sized array for which macros automatically implement traits
+    // Until serde supports const generics we need to use serde_big_array for fixed size arrays
+    // Larger than 32.
+    const MAX_FIXED_ARRAY_LEN: usize = 32;
+    let serde_line = match field.field_type.array_info {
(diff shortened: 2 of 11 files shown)


Carter12s:
@ssnover Do you have any good ideas on how we should handle this situtation?

1. Is there a workaround where maybe we can use some other type than [T; N] to represent a fixed sized array?
2. Can we codegen our own trait impls for larger array sizes? (This seems awful)
3. Do we just declare our crate only works with fixed sized arrays <32 elements until const generics land? can we limit that impact to ROS1 usage only? probably not cause it affects codegen...

Am I just missing an obvious solve here.

ssnover:
We have at least one instance showing in the codegen for a 36 member array, so we probably ought to at least support that as a basic ROS message. 

Only alternative I can think of is to represent with a tuple of Vec and the fixed size and then define our own serialization and deserialization for serde_rosmsg?

Carter12s:
I have a little side test setup in my fork of serde_rosmsg, from when the bug was originally reported, where I'm deserializing a NavSatFix.msg with the library. NavSatFix has an [f64; 9] in it.

For Box<[f64]> serde_rosmsg spat out an underflow error at runtime.
For &[f64] we get a compile time "Deserialize<'_> is not implemented"

So don't see a good path there...
I may open an issue on serde_rosmsg and see if they have a solve...

Carter12s:
https://stackoverflow.com/questions/62665558/how-can-i-implement-serdedeserialize-for-arrays-larger-than-32

Has lead me to: https://crates.io/crates/serde-big-array

Which I'm going to play with and see if it can solve this problem...

Carter12s:
So it looks like a combination of some fuckery with smart_default's _code directive and this serde-big-array crate I may have gotten a working solution.

Biggest problem I see is 1) the code is super messy and could use some unit tests, and 2) we're leaking yet another crate dependency through the codegen interface in not a great way...

I'd love your eyes on this @ssnover

roslibrust_codegen/src/gen.rs
```diff
@@ -145,20 +147,48 @@ fn generate_field_definition(
...
+                let default_str = format!("[Default::default(); {fixed_array_length}]");
+                quote! { #[default(_code = #default_str)]}
+            } else {
+                quote! {}
+            }
+        } else {
+            quote! {}
+        }
+    };
+    let serde_line = if let Some(Some(fixed_array_length)) = field.field_type.array_info {
```
ssnover, line 176: Nitpick: This is a little unclean looking, perhaps:
```rust
let serde_line = match field.field_type.array_info {
    Some(Some(fixed_array_len)) if fixed_array_len > MAX_FIXED_ARRAY_LEN => quote! { #[serde(with = "::serde_big_array::BigArray")] },
    _ => quote! {},
};
```

roslibrust_codegen/src/parse/mod.rs
```diff
@@ -216,3 +216,24 @@ fn parse_type(type_str: &str, pkg: &Package) -> Result<FieldType, Error> {
         }
     }
 }
+
+#[cfg(test)]
+mod test {
```
ssnover, line 221: Big fan of introducing more test cases like this.

roslibrust_codegen/src/lib.rs
```diff
@@ -631,7 +631,7 @@ fn parse_ros_files(
 
 #[cfg(test)]
 mod test {
-    use crate::find_and_generate_ros_messages;
+    use crate::{find_and_generate_ros_messages, FieldInfo};
```
ssnover, line 634: Looks like this was unused.

roslibrust/Cargo.toml
```diff
@@ -50,6 +50,7 @@ regex = { version = "1.9", optional = true } # Only used with native ros1
 env_logger = "0.10"
 test-log = "0.2"
 simple_logger = "2.1.0"
+serde-big-array = "0.5"
```
ssnover, line 53: This was not picked up in the noetic CI. This is the only blocker I think.

ssnover (changes requested):
This is great, thanks for taking the time to figure out the attribute macro magic required to get this functioning and adding tests.

I pointed out the current CI failures in comments.

roslibrust_codegen/src/gen.rs
Carter12s: Love that much cleaner.

Issue #165, opened 2024-06-29
New ROS1 Service Server, Does Clean-up on Drop


Carter12s:
Service server should clean-up on drop this was left unimplemented in code as a todo and needs to get implemented.

PR #168, opened 2024-07-01
Fix Service Servers Shutdown Cleanly

Carter12s:
## Description
Add logic to both ServiceClient and ServiceServer to correctly clean-up after themselves when their respective handles are dropped.

## Fixes
Closes: #165 

## Checklist
- [ ] Update CHANGELOG.md


Carter12s:
Service Server logic is in place in MR, however in testing discovered that ServiceClient logic is also broken.

Going to fix in this same PR.

Issue #173, opened 2024-07-06
Plotjuggler 'Missing ROSType in library' on ros1 node published messages


lucasw:
Other nodes get the messages fine, but plotjuggler is trying to get type information in a way roslibrust nodes don't provide?

https://github.com/facontidavide/PlotJuggler/blob/main/plotjuggler_plugins/ParserROS/rosx_introspection/src/ros_message.cpp#L206

What I do is click `Streaming | ROS Topic Subscriber | Start' within plotjuggler, select the topic and press OK, and then get the error:

![plotjuggler_roslibrust_missing_datatype](https://github.com/Carter12s/roslibrust/assets/1334122/7b664ffd-bc7c-44d4-8292-55457c073140)

I can start up a roscpp/rospy node publishing to the same topic, get plotjuggler plotting it, then switch in the rust node and it'll still plot fine (presumably having already cached the type information it wanted).

This happened with an earlier version of roslibrust (883057f8d159b0298c243744d65971e746ec5dbc) and then I just now updated to the latest in master with same result.

Carter12s:
Lucas thank you!!! Seriously, awesome to have an active user, and really appreciate these issues. I'm just using this crate in my own little world and its awesome to have someone catching these things.

I haven't used plot juggler before, but I'm aware of it, let me dig in and see if this will be an easy fix.

Carter12s:
Okay after doing some digging.

Plot Juggler ultimately parses the message_definition= field of the TCPROS connection header (which we should be providing) via RosIntrospection::ShapeShifter [here](https://github.com/PlotJuggler/plotjuggler-ros-plugins/blob/d0692ad5fb7ea9f314f40816ea9e0a4fe55f8397/src/DataStreamROS/datastream_ROS.cpp#L64C1-L75C56).

So my first guess / assumption is that RosIntrospection::ShapeShifter isn't giving them the message definition from us.

Carter12s:
Well crap... turns out I was wrong... ROS does have the message_definition field in the connection header contain the fully expanded definition...

Apparently that field is supposed to contain the full output of `gendeps --cat` example here:
https://wiki.ros.org/roslib/gentools

Which is not what our current message generation is doing :poop: 100% a bug. Will be somewhat annoying to fix.

Carter12s:
For my own reference when I dig into implementation, gendeps concatenation is dead simple:

[...]

PR #174, opened 2024-07-06
Fixed code generation to include fully expanded message definition

Carter12s:
## Description
Previously our message_definition provided by Publishers did not include the fully expanded definition. This works fine for most cases, but PlotJuggler was fully parsing the definition for dynamic interaction. This should get our message_definition to be accepted by PlotJuggler.

## Fixes
Closes: #173 

## Checklist
- [x] Update CHANGELOG.md


lucasw:
With rostopic pub I see this from `repr(msg._connection_header["message_definition"])` in a PointStamped:

```
rostopic pub /point geometry_msgs/PointStamped "{point: {x: 1.0}}"
```

```
"# This represents a Point with reference coordinate frame and timestamp\nHeader header\nPoint point\n\n================================================================================\nMSG: std_msgs/Header\n# Standard metadata for higher-level stamped data types.\n# This is generally used to communicate timestamped data \n# in a particular coordinate frame.\n# \n# sequence ID: consecutively increasing ID \nuint32 seq\n#Two-integer timestamp that is expressed as:\n# * stamp.sec: seconds (stamp_secs) since epoch (in Python the variable is called 'secs')\n# * stamp.nsec: nanoseconds since stamp_secs (in Python the variable is called 'nsecs')\n# time-handling sugar is provided by the client library\ntime stamp\n#Frame this data is associated with\nstring frame_id\n\n================================================================================\nMSG: geometry_msgs/Point\n# This contains the position of a point in free space\nfloat64 x\nfloat64 y\nfloat64 z\n"
```

but the from a roslibrust node using this branch (at 03d73ac6)) it's still not expanding the other fields:

```
'# This represents a Point with reference coordinate frame and timestamp\nHeader header\nPoint point'
```

Carter12s:
Okay, was able to get to an actual dev environment today where I could directly test this. Found a bug in how I was collecting referenced types that was skipping a layer.

* Manually confirmed works with plot juggler now
* Manually confirmed works with rospy
* Added at least a basic unit test to confirm message generation is generating the expanded types into the future

image

Issue #190, opened 2024-08-06
Subscribe to any topic type like ShapeShifter/AnyMsg and get raw bytes from next() instead of deserializing


lucasw:
I made a subscribe-to-anything system here: https://github.com/lucasw/roslibrust/tree/subscribe_any (and https://github.com/lucasw/mcap_tools/blob/main/mcap_tools/src/mcap_record.rs makes use of it)- it tries to duplicate the rospy/roscpp approach but also requires a placeholder ros message type to satisfy other code.

Could there be a special ShapeShifter/AnyMsg completely internal to roslibrust that satisfies all the traits of a ros message type as generated by the macro, and even can get passed into serde_rosmsg and just returns Vec<u8> (or returns what seems like a regular message but contains only `byte[] data`).

Maybe something is wrong with my implementation as I get duplicate received messages if the mcap_record node starts before the publisher.

What I'm looking into next is getting information out of the connection header to satisfy writing an mcap record (mostly getting the message definition), looks like I can find the right `Subscription` in the node handle and access it.

Further future is the publisher equivalent of this (for a rosbag play alike)- be able to give the publisher raw bytes to publish after configuring it with the topic and message definition and md5sum loaded from the mcap.

I haven't tried rebasing on your latest changes, but one thing that stands out is the part you commented out that compares md5sums.

Some notes on where I found the values rospy and roscpp were using: https://github.com/lucasw/mcap_tools/issues/1#issuecomment-2271952018

Carter12s:
Okay, spent some time thinking on this today.

1. I think this is a useful feature for the crate to have and pretty easy to support so we should add it.

2. I think this should happen through a parallel API to subscribe() (much like you've done) and not rely on any "Magic Type"
  a. Call to subscribe_raw() -> returns a Subscriber<Vec<u8>>. We don't need subscribe_raw() to take any type argument.
  b. Instead of Subscriber having the deserialization logic buried inside .next(), Subscriber holds a "Fn(Vec<u8>) -> T". This adds an extra layer of indirection and there will be some runtime performance hit, but not trying to make this crate stupid fast.
  c. When Subscriber gets constructed either in .subscribe() or in .subscribe_raw() the correct handler function is inserted into it. For a regular subscribe this will perform the deserialization logic. For a raw subscribe this can be a no-op and return the raw Vec<u8>

[...]

PR #192, opened 2024-08-10
subscribe to any topic of any type and get received messages as a raw vector of bytes

lucasw:
## Description

Subscribe to a topic without needing to know the type, and then get the received message as a raw vector of bytes, similar to rospy AnyMsg and roscpp ShapeShifter.

This isn't quite what was suggested in https://github.com/Carter12s/roslibrust/issues/190#issuecomment-2276521772-  I couldn't figure out how to make a function within the existing Subscriber that can do different things depending on type, it needs specialization?  Is having both a Subscriber and SubscriberAny bad for other parts of roslibrust?

## Fixes
Closes: #190

## Checklist
- [ :heavy_check_mark: ] Update CHANGELOG.md

### DIFF
--- a/roslibrust/src/ros1/subscriber.rs
+++ b/roslibrust/src/ros1/subscriber.rs
@@ -1,6 +1,6 @@
 use crate::ros1::{names::Name, tcpros::ConnectionHeader};
 use abort_on_drop::ChildTask;
-use roslibrust_codegen::RosMessageType;
+use roslibrust_codegen::{RosMessageType, ShapeShifter};
 use std::{marker::PhantomData, sync::Arc};
 use tokio::{
     io::AsyncWriteExt,
@@ -41,7 +41,7 @@ impl<T: RosMessageType> Subscriber<T> {
 
 pub struct SubscriberAny {
     receiver: broadcast::Receiver<Vec<u8>>,
-    _phantom: PhantomData<Vec<u8>>,
+    _phantom: PhantomData<ShapeShifter>,
 }
 
 impl SubscriberAny {
@@ -52,6 +52,7 @@ impl SubscriberAny {
         }
     }
 
+    // pub async fn next(&mut self) -> Option<Result<ShapeShifter, SubscriberError>> {
     pub async fn next(&mut self) -> Option<Result<Vec<u8>, SubscriberError>> {
         let data = match self.receiver.recv().await {
             Ok(v) => v,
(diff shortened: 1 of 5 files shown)


roslibrust/src/ros1/subscriber.rs
```diff
@@ -39,6 +39,29 @@ impl<T: RosMessageType> Subscriber<T> {
     }
 }
 
+pub struct SubscriberAny {
+    receiver: broadcast::Receiver<Vec<u8>>,
+    _phantom: PhantomData<Vec<u8>>,
+}
+
+impl SubscriberAny {
```
Carter12s, line 47: If we go with this approach, I think we also need to implement Drop for SubscriberAny matching what Subscriber::drop() does to un-subscribe under the hood.

Carter12s: May be worth duplicating the subscriber dropping integration test for SubscriberAny to make sure behavior stays the same.

roslibrust_codegen/src/integral_types.rs
```diff
@@ -36,6 +36,13 @@ impl RosMessageType for Time {
     const DEFINITION: &'static str = "";
 }
 
+// The equivalent of rospy AnyMsg or C++ ShapeShifter, subscribe_any() uses this type
+impl RosMessageType for Vec<u8> {
```
Carter12s, line 40: I think we probably want to do a "newtype" pattern here, and say `type RawBytesMsg(Vec<u8>)` (ShapeShifter might be a better name for it).

And then `impl RosMessageType for RawBytesMsg` instead of for Vec<u8> directly?

Not a huge difference, but I am worried about what happens if a crate user calls `.subscribe::<Vec<u8>()` with this definition floating around? With the `newtype` pattern we can make sure "RawBytesMsg" is a private type of the crate and not available for the user to try to use somewhere else? Maybe doesn't matter?

Carter12s:
I'm not opposed to this as the solution, and I think having this in the crate is better than not having anything, so I'm okay with merging this.

The part where this can get messy is with the TopicProvider trait I'm working on now. Having SubscriberAny just makes it hard (not impossible) to create a generic trait that represents a ROS api. It adds another generic associated type on the trait, but that is probably fine?

If you are interested I'd be happy to throw together a version of what I was think of, and we could compare them?

Otherwise happy to merge this as is (once Drop stuff is fixed).

roslibrust/src/ros1/subscriber.rs
Carter12s: Whoops sorry, lost my mind for a second!

For ros1, dropping the last tx side of the channel causes subscription clean-up. So this is good!

roslibrust_codegen/src/integral_types.rs
Carter12s: What you did is perfect!

Carter12s:
If you have publisher any ready, go ahead and put PR up. TopicProvider stuff can take a back seat!

Issue #206, opened 2024-11-16
remote ros1 subscriber disconnect causes publisher to hang, doesn't publish to other subscribers that are still connected


lucasw:
I think this is just a specific case of #14

Modify ros1_talker to publisher at a much higher rate in a loop instead of a 1 second delay for only 50 counts:

```
    // for count in 0..50 {
    let mut count = 0;
    loop {
        let mut msg = geometry_msgs::PointStamped::default();
        msg.header.seq = count;
        msg.point.x = count as f64;
        publisher.publish(&msg).await?;
        tokio::time::sleep(tokio::time::Duration::from_millis(10)).await;
        count += 1;
    }
```

Start a roscore and talker on one computer:

```
roscore
```
```
ROS_IP=master_ip cargo run --release --example ros1_talker
```

```
rostopic echo /my_point
```

Then on another computer on the same network:

```
ROS_MASTER_URI=http://master_ip:11311 ROS_IP=remote_computer_ip rostopic echo /my_point
```

After letting it go a while disconnect the remote computer- close the laptop lid if that causes the computer to sleep, or disconnect via network manager, or pull the ethernet cable, or similar.  Obviously the remote rostopic echo stops working immediately, but the issue is the echo on the same computer as the publisher stops working after a few seconds also (9 seconds in one test, 900 message after the disconnect given the 100 Hz update rate).  Reconnecting usually recovers both echos quickly.

Is this await blocking the loop to send data to the rest of the subscribers? 
 Maybe there are options to make it time out quickly? https://github.com/RosLibRust/roslibrust/blob/master/roslibrust/src/ros1/publisher.rs#L217
I can imagine a restructure where there's a broadcast channel with limited queue size sending messages (ideally not cloned) to per subscriber tokio tasks, and then if one hangs it won't bring down the others, but maybe there's a quicker fix.

This is all tested on Ubuntu 22.04/24.04.

Carter12s:
Spent some time looking at this yesterday and today.

# Testing Rant

This really deserves a unit / integration test, but it ends up being remarkably hard to write one. Emulating a TCP socket disconnecting mid test is proving challenging to replicate:

- `tc netem` is hard to setup inside a docker image, couldn't figure out kernel permissions here
- [Tokio Turmoil](https://github.com/tokio-rs/turmoil) seems like it would do exactly what we want, but we can't directly use it cleanly as we want "real" communication with rosmaster while mocking communication between nodes.
[...]

PR #208, opened 2024-11-25
Working publisher using broadcast

Carter12s:
## Description
Re-working publisher to use tokio::broadcast as the underlying channel so that one lagging TCP Socket doesn't result in the publisher blocking.

## Fixes
Closes: #206 

## Checklist
- [ ] Update CHANGELOG.md

### DIFF
--- a/roslibrust/src/ros1/node/actor.rs
+++ b/roslibrust/src/ros1/node/actor.rs
@@ -44,7 +44,7 @@ pub enum NodeMsg {
     // This results in the node's task ending and the node being dropped.
     Shutdown,
     RegisterPublisher {
-        reply: oneshot::Sender<Result<broadcast::Sender<Vec<u8>>, String>>,
+        reply: oneshot::Sender<Result<(broadcast::Sender<Vec<u8>>, mpsc::Sender<()>), String>>,
         topic: String,
         topic_type: String,
         queue_size: usize,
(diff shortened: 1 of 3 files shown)


lucasw:
I can test this more in a few days- but did I try removing the publish async: https://github.com/lucasw/roslibrust/commit/395cf9949f19e62c3e898871444e5e0e6490d43f - want to have that change here (and then deal with all the downstream pubilsh() changes)?

lucasw:
I can test this more in a few days- but did I try removing the publish async: https://github.com/lucasw/roslibrust/commit/395cf9949f19e62c3e898871444e5e0e6490d43f - want to have that change here (and then deal with all the downstream pubilsh() changes)?

Carter12s:
I'd like to avoid the `removing async from publish` change in this MR. I don't think it is a crazy change to make, but I think one change at a time.

lucasw:
One thing I notice testing this is that if the remote `rostopic echo` is started before the local publisher, it never makes the connection and doesn't echo anything unless restarted, while a local rostopic echo works fine.  This appears to be true both before and after the changes here (I'll try jumping back more commits and see if there was an earlier regression).

Besides that the intended fix here looks good, and the connection even recovers in the case of disconnecting ethernet, waiting a minute, then reconnecting (but too much longer gets more fully disconnected or it's just hung up forever and even reconnections don't recover it?).

Carter12s:
I went ahead and merged with your testing cause I like these changes, but definitely more work to do here to make ros1 robust over various disconnection scenarios. Lots of complexity buried in there.

PR #215, opened 2024-12-10
Major re-org to crate / repository structure

Carter12s:
## Description
A major re-organization of the components of roslibrust per: https://github.com/orgs/RosLibRust/discussions/214

Some facets of this:
 - The core roslibrust crate now includes no backends by default. The rosbridge backend is now hidden behind the rosbridge feature flag. This means you don't pay to compile any backend that you aren't using.
 - The rosbridge backend moved under the rosbridge namespace in the roslibrust crate so it is no longer treated as a special backend in the naming conventions.
 - RosApi trait and implementations moved out of roslibrust and into their own crate. Also now generically written so they can be used with any backend.

## Fixes
Closes: `number`

## Checklist
- [ ] Update CHANGELOG.md


Carter12s:
@lucasw / @ssnover this is ready for a review now.

It is a pretty big change to overall crate structure, but I think aligns things much more where I want development to head.

I think it sets things up well to eventually include a ros2 backend with either zenoh or dds, and I think it sets us up very well to start making the generic traits for TopicProvider and ServiceProvider the a first class API that both allows folks to swap backends around and to use MockRos for testing.

ssnover:
Reading on mobile right now, but just going to ask clarifying questions and I'll review later:

* Is the intended entry point crate roslibrust_common now?
* Should it maybe be called roslibrust_core?

Carter12s:
For users the intended entrypoint is still roslibrust

The "standard" usage would be to depend on roslibrust with at least one backend enabled via a feature flag.

Depending on roslibrust with no backends enabled can be done (see rosapi) to implement generic behaviors without a particular backend specified as well.

Not expecting anyone outside of roslibrust development to depend on roslibrust_common and I'm super happy to rename that to roslibrust_core or potentially `roslibrust_traits` <- that is most specifically what is actually inside it at the moment.

roslibrust/Cargo.toml
```diff
@@ -6,82 +6,49 @@ edition = "2021"
...
 
 [features]
 default = []
 # Note: all does not include running_bridge as that is only intended for CI
 all = []
 # Provides a rosapi rust interface
-rosapi = ["serde-big-array"]
+# TODO MAJOR remove this feature
+rosapi = []
 # Intended for use with tests, includes tests that rely on a locally running rosbridge
```
Carter12s, line 38: Flagging for myself I still need to clean this up before merge

example_package/Cargo.toml
```diff
@@ -6,6 +6,8 @@ edition = "2021"
 [dependencies]
 # The code generated by roslibrust_codegen has dependendencies
 # We need to depend on the crate at build time so that the generate code has access to these dependencies
+roslibrust_common = { path = "../roslibrust_common" }
+# THIS SHOULDN'T BE NEEDED, BUT IS FOR NOW GOTTA FIX LEAKING DEPENDENCIES
```
ssnover, line 10: Can you make a tracking issue for this?

roslibrust/Cargo.toml
```diff
@@ -6,82 +6,49 @@ edition = "2021"
...
-serde_xmlrpc = { version = "0.2", optional = true } # Only used with native ros1
-roslibrust_serde_rosmsg = { version = "0.4", optional = true } # Only used with native ros1
-hyper = { version = "0.14", features = [
-    "server",
-], optional = true } # Only used with native ros1
-gethostname = { version = "0.4", optional = true } # Only used with native ros1
-regex = { version = "1.9", optional = true } # Only used with native ros1
-# TODO I think we should move rosapi into its own crate...
-serde-big-array = { version = "0.5", optional = true } # Only used with rosapi
+# TODO version here
```
ssnover, line 14: TODO is for this PR or a latter one?

roslibrust/Cargo.toml
```diff
@@ -6,82 +6,49 @@ edition = "2021"
...
+roslibrust_mock = { path = "../roslibrust_mock", optional = true }
 
 [dev-dependencies]
-env_logger = "0.10"
+env_logger = "0.11"
 test-log = "0.2"
-simple_logger = "5.0"
-serde-big-array = "0.5"
+log = "0.4"
+tokio = { version = "1.20", features = ["macros", "rt-multi-thread"] }
```
ssnover, line 25: Since it's a dev-dependency it probably could just be "1" for the version.

roslibrust/Cargo.toml
```diff
@@ -6,82 +6,49 @@ edition = "2021"
...
+serde = { version = "1.0", features = ["derive"] }
```
ssnover, line 26: Same versioning argument for serde.

roslibrust/Cargo.toml
```diff
@@ -6,82 +6,49 @@ edition = "2021"
...
+roslibrust_codegen = { path = "../roslibrust_codegen" }
+roslibrust_codegen_macro = { path = "../roslibrust_codegen_macro" }
 
 [features]
 default = []
 # Note: all does not include running_bridge as that is only intended for CI
 all = []
 # Provides a rosapi rust interface
-rosapi = ["serde-big-array"]
+# TODO MAJOR remove this feature
```
ssnover, line 36: Tracking issue?

roslibrust/src/tests/ros1_native_integration_tests.rs
```diff
@@ -4,7 +4,7 @@
 #[cfg(feature = "ros1_test")]
 mod tests {
     use log::*;
-    use roslibrust::ros1::{NodeError, NodeHandle};
+    use roslibrust_ros1::{NodeError, NodeHandle};
```
ssnover, line 7: Since integration tests will have similar visibility to internals as user code, I think it's probably best for the integration test to look like what we expect it to look like for users; i.e. keep `roslibrust::ros1::`

roslibrust_common/src/lib.rs
```diff
@@ -0,0 +1,116 @@
...
+    // Generic catch-all error type for not-yet-handled errors
+    // TODO ultimately this type will be removed from API of library
+    #[error(transparent)]
+    Unexpected(#[from] anyhow::Error),
+}
+
+/// Generic result type used as standard throughout library.
+/// Note: many functions which currently return this will be updated to provide specific error
+/// types in the future instead of the generic error here.
+pub type RosLibRustResult<T> = Result<T, RosLibRustError>;
```
ssnover, line 35: While you're making breaking changes, it'd be great to rename this to `Result` and let namespacing be `roslibrust::Result`. Ditto for `RosLibRustError`.

roslibrust_ros1/src/node/actor.rs
```diff
@@ -197,7 +194,7 @@ impl NodeServerHandle {
 
         let md5sum;
         let md5sum_res =
-            roslibrust_codegen::message_definition_to_md5sum(topic_type, msg_definition);
+            roslibrust_common::md5sum::message_definition_to_md5sum(topic_type, msg_definition);
         match md5sum_res {
```
ssnover, line 198: This is an aside, but perhaps:
```rust
let md5sum = roslibrust_common::md5sum::message_definition_to_md5sum(
    topic_type,
    msg_definition
).map_err(|err| {
    log::error!("{err:?}");
    Err(NodeError::IoError(io::ErrorKind::ConnectionAborted.into()))
})?;
```

ssnover (commented):
Mostly nitpicks and questions about tracking todos.

roslibrust/src/tests/ros1_native_integration_tests.rs
ssnover: Oh looking at this now, I see this is under `src` instead of `tests` which is where integration tests ought to go (since they have the same visibility into the crate as a user of the crate does).

roslibrust_ros1/src/node/actor.rs
ssnover: It might be nice to change the name of `message_definition_to_md5sum` to `from_message_definition` since it's already namespaced under `md5sum` module.

roslibrust/Cargo.toml
Carter12s: Probably for a later one?

I'll make a tracking issue for the v0.12 release which will be a much larger one than normal.

Gotta update all the release instructions and flows with the new crates.

roslibrust/Cargo.toml
Carter12s: Good call... I actually think I'll just go ahead and make tokio a workspace dependency and switch everything to using that. We have way to many versions floating around.

roslibrust/src/tests/ros1_native_integration_tests.rs
Carter12s: Good catch I'll re-org!

roslibrust_common/src/lib.rs
Carter12s: AGREED

roslibrust/examples/basic_publisher.rs
```diff
@@ -1,19 +1,16 @@
...
 /// To run this example a rosbridge websocket server should be running at the deafult port (9090).
+#[cfg(feature = "rosbridge")]
 #[tokio::main(flavor = "multi_thread")]
-async fn main() -> Result<(), anyhow::Error> {
-    simple_logger::SimpleLogger::new()
-        .with_level(log::LevelFilter::Debug)
-        .without_timestamps() // required for running wsl2
-        .init()
-        .unwrap();
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
```
lucasw, line 10: Why the change from `anyhow`- just to get rid of the dependency?

roslibrust/examples/basic_publisher.rs
```diff
@@ -1,19 +1,16 @@
...
+    use log::*;
+    use roslibrust::rosbridge::ClientHandle;
+    env_logger::init();
```
lucasw, line 13: Similar question here, though it looks like you replaced simple_logger with env_logger everywhere- is env_logger better?

roslibrust/examples/basic_publisher.rs
Carter12s: Yeah mainly just trying to simplify examples and use the least amount of extra things in them.

roslibrust/examples/basic_publisher.rs
Carter12s: I've found env_logger to the most common "starting out logger."

I can't remember exactly why I started out with simple_logger, but subsequently have found env_logger to just be a lot more commonly used, so this seemed like a slightly "happier path" for the examples.

Carter12s: I just did a check since I was curious:

7.5M downloads per month: https://lib.rs/crates/env_logger
1M downloads per month: https://lib.rs/crates/simple_logger

So it isn't like simple_logger isn't used, but env_logger is ~7x more used.

ssnover: env_logger is more flexible to configure but has the common pitfall of not piping to stdout/stderr unless you specify RUST_LOG env variable and some people find that as a hurdle for simple logging.

example_package/Cargo.toml
Carter12s: https://github.com/RosLibRust/roslibrust/issues/216

roslibrust/Cargo.toml
Carter12s: https://github.com/RosLibRust/roslibrust/issues/217

Issue #229, opened 2025-04-17
Commented out python example code within .msg/.srv files causes find_and_generate_ros_messages() errors


lucasw:
If I have mavros_msgs on my ROS_PACKAGE_PATH when building `example_package_macro` I get this:

```
error: expected one of `.`, `;`, `?`, `where`, or an operator, found `" + ename)
       #     for k, e in enum:
       #         sn = e.name[len(ename) + 1:]
       #         l = "uint`
  --> example_package_macro/src/main.rs:8:1
   |
8  | / roslibrust::find_and_generate_ros_messages!(
9  | |     "assets/ros1_test_msgs",
10 | |     "assets/ros1_common_interfaces"
11 | | );
   | | ^
   | | |
   | | expected one of `.`, `;`, `?`, `where`, or an operator
   | |_while parsing this item list starting here
   |   the item list ends here
   |
   = note: this error originates in the macro `roslibrust::find_and_generate_ros_messages` (in Nightly builds, run with -Z macro-backtrace for more info)
```

Copy https://github.com/mavlink/mavros/blob/master/mavros_msgs/msg/ADSBVehicle.msg into assets/ros1_test_msgs/msg and build `example_package` it errors differently:

```
error: expected `;`, found `" + ename)
       #     for k, ev in enum:
       #         // sn = ev.name[len(ename) + 1:]
       #         l = "uint`
   --> /home/lucasw/rust/misc/roslibrust/target/debug/build/example_package-9661660fac5f3a9a/out/messages.rs:521:18
    |
521 |   #     cog.outl("# " + ename)
    |  __________________^-
    | |                  |
    | |                  help: add `;` here
522 | | #     for k, ev in enum:
523 | | #         // sn = ev.name[len(ename) + 1:]
524 | | #         l = "uint { bsz } { pfx } { sn } = { k } ".format(**locals())
    | |___________________- unexpected token

error: non-item in item list
   --> /home/lucasw/rust/misc/roslibrust/target/debug/build/example_package-9661660fac5f3a9a/out/messages.rs:521:19
    |
[...]

PR #230, opened 2025-04-17
add more hashes to raw string literal to avoid unusual ros message definition comment problems

lucasw:
## Description

Messages like https://github.com/mavlink/mavros/blob/master/mavros_msgs/msg/ADSBVehicle.msg fail to build as described in #229 without this raw string literal change- I've added a copy of it into `ros1_test_msgs/`.

## Fixes
Closes: #229

## Checklist
- [x] Update CHANGELOG.md


Carter12s:
Awesome work! Good find!

PR #232, opened 2025-06-11
Handling case where a # is present in a string constant field

adrienbarral:
## Description
By very carefully reading the ros 1 documentation (and of course, encounter an issue, that prompted me reading the documentation 😏 ), I understand that a # character in a string constant should not be interpreted as a comment.

See this small sentence [here in the doc](https://wiki.ros.org/msg#:~:text=string%20constants%20are-,assigned,-the%20value%20of)

That means the following .msg file : 
`string FOO=foo # bar` 
must product a constant with the value "foo # bar".

I can't find the same level of specification in ROS2, but I guess the same behaviour is required.

So, my PR is about fixing this particular case.

### DIFF
--- a/roslibrust_codegen/src/parse/mod.rs
+++ b/roslibrust_codegen/src/parse/mod.rs
@@ -108,7 +108,7 @@ fn parse_field(line: &str, pkg: &Package, msg_name: &str) -> Result<FieldInfo, E
     })
 }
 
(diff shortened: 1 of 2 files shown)


adrienbarral:
I did a small mistake (and I will correct it very soon). I did the assumption that String and string are the same type for constant. This is not true. So the inherent rust type of a String must be a `&'static str` and rust type for string must remain a std::String...

I will add a else if branch in the mod.rs line 128.

Sorry.

PR #273, opened 2025-11-13
refactor: export the message file interface

YuanYuYuan:
I propose exporting the message file interface so we can leverage this in ros-z. BTW, [this](https://github.com/ZettaScaleLabs/ros-z/pull/11) is an PoC of ROS-Z using ROS messages generated by roslibrust!


Carter12s:
Happy with this and fully support it!

Excited about your PoC!

I see you doing some cool extensions to roslibrust_codegen in ros-z-codegen. Let me know if you think we can/should merge those capabilities into roslibrust_codegen. I'd really a situation where users could generate types once in a consistent way and those type would work with either roslibrust or ros-z without having to jump through any hoops.

PR #289, opened 2025-12-02
[ROS1/TCPROS] Improve self-ip heuristics.

0x53A:
When neither ROS_IP nor ROS_HOSTNAME are set, use a slightly smarter heuristic to try to find an IP address in the same subnet as the rosmaster. This only works if the rosmaster is specified as an IP address, not if it is a hostname.

closes https://github.com/RosLibRust/roslibrust/issues/288

_(opened as Draft so I can test it in my own system)_

### DIFF
--- a/roslibrust_ros1/Cargo.toml
+++ b/roslibrust_ros1/Cargo.toml
@@ -32,6 +32,7 @@ regex = { version = "1.9" }
 byteorder = "1.4"
 thiserror = "2.0"
 anyhow = "1.0"
+if-addrs = "0.14.0"
 
 [dev-dependencies]
 # Used for message definitions in tests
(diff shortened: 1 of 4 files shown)


roslibrust_ros1/Cargo.toml
```diff
@@ -32,6 +32,7 @@ regex = { version = "1.9" }
 byteorder = "1.4"
 thiserror = "2.0"
 anyhow = "1.0"
+if-addrs = "0.14.0"
```
Carter12s: `get_if_addrs` is explicitly a non-windows function in the if_addrs crate. Which means adding this MR explicitly makes `roslibrust` not support windows. While we haven't been explicitly testing for or advertising windows support, I actually do expect it to work across all our backends at the moment.

Looking around it looks like the `getifs` crate may be a better alternative to grab that does support windows?

Issue #296, opened 2025-12-30
Incorrect Behavior When Advertising Without Initial /


Carter12s:
Bigger issue here is just inconsistency in how / is either prepended or not.

And a little bit our not supporting inherited name spacing with ~

Goal:

- For all backends have .advertise("my_topic") and .advertise("/my_topic") behave identically and connect to eachother
- Have unit tests for all back-ends for this

PR #299, opened 2026-01-12
Fix topic names flexibility

Carter12s:
## Description
Major overhaul of how topic names are using throughout the public API.

## Fixes
Closes: #296 

## Checklist
- [x] Update CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- Reorganized some of the module structure of roslibrust_common. User's pulling things from `roslibrust` will see no change.
+- Reorganized some of the module structure of roslibrust_common. Users pulling things from `roslibrust` will see no change.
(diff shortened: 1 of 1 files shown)


roslibrust_common/src/topic_name.rs
```diff
@@ -0,0 +1,183 @@
...
+}
+
+impl ToGlobalTopicName for &str {
+    fn to_global_name(self) -> Result<GlobalTopicName, RError> {
+        GlobalTopicName::new(self)
+    }
+}
+
+static GLOBAL_NAME_REGEX: std::sync::LazyLock<regex::Regex> = std::sync::LazyLock::new(|| {
+    regex::Regex::new(r"^([/~a-zA-Z]){1}([a-zA-Z0-9_/])*([A-z0-9_])$").unwrap()
```
tazlor: In the last capture group, `A-z` includes non-alphanumeric characters that I don't think you intend to capture.
And this is allowing a leading non-slash character, so it's not actually enforcing roslibrust's requirements. What's up with that?

PR #303, opened 2026-01-14
Optimize ros1

Carter12s:
## Description
Did an profile guided optimization pass on the ROS1 backend. Found ~79% speedup via reducing copies (as measured on 1080p image benchmark). The bulk of the improvement comes from swapping Vec<u8> for Bytes which is equivalent to Arc<Vec<u8>> so that fewer clones of the underlying data were made throughout handling. Additionally, some buffer shenanigans were cleaned up to prevent redundant allocations and moves when trying to append the length of the message at the front. 

## Fixes
Closes: `number`

## Checklist
- [x] Update CHANGELOG.md


Carter12s:
Flamegraph of image_bench from master branch:

master_flamegraph

Flamgraph from this state:
v2_flamegraph

PR #323, opened 2026-05-12
Cleanup ros1 Subscriber subscription on drop

DomWilliamsEE:
## Description
Subscriber `Drop` impl now unregisters the subscription with master and the publishing node if it's the last subscriber alive. Based on the discussion from 2024 in #200, possibly there might be a better method preferred now. Manual testing shows that it works as expected without leaking resources.

Disclosure: Code changes are Claude-assisted :robot: but manually reviewed and tested.

## Fixes
Closes: #200 

## Checklist
- [ ] Update CHANGELOG.md

### DIFF
--- a/roslibrust_ros1/src/node/actor.rs
+++ b/roslibrust_ros1/src/node/actor.rs
@@ -810,10 +810,13 @@ impl Node {
             return Ok(());
         }
         let _ = self.subscriptions.remove(topic);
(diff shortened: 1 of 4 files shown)


Carter12s:
Awesome! Thank you for this contribution.

I'm going to extend this merge request with a specific unit/integration test to cover this case to make sure we don't regress here, but overall this looks solid.

PR #325, opened 2026-05-16
Optimize ros1

Carter12s:
## Description
- Provides a ~20-30% speedup (high variance) on image benchmark by doing more intelligent pre-allocation

This may have some performance penalties on topics which have frequent small messages and infrequent large messages, but I'm not really sure we'll see a real-world case of that.

## Fixes
Closes: `number`

## Checklist
- [ ] Update CHANGELOG.md

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -70,6 +70,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "aligned-vec"
+version = "0.6.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "dc890384c8602f339876ded803c97ad529f3842aba97f6392b3dba0dd171769b"
+dependencies = [
+ "equator",
+]
+
+[[package]]
 name = "allocator-api2"
 version = "0.2.21"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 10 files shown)


Carter12s:
image

This is a worst-case performance improvement on the image bench, I've done a lot of runs and see high std.dev between runs, but with 10k samples over ~15s I think this is pretty real.

Carter12s:
I've also generated an equivalent C++ benchmark and run it on the same system the above benchmarking was done on. I've done my best to make the benchmarks as equivalent as possible:

```
rosrun roslibrust_cpp_image_bench cpp_image_bench --mode driver --iterations 10000 --warmup 50
mode: driver
samples: 10000 measured, 50 warmup
image: 1920x1080 rgb8, payload_bytes: 6220800
mean: 1.228 ms
median: 1.203 ms
p95: 1.377 ms
min: 0.992 ms
max: 4.984 ms
stddev: 0.094 ms
throughput_payload: 4832.606 MiB/s (5.067 GB/s)
json: {"mode":"driver","iterations":10000,"warmup":50,"width":1920,"height":1080,"channels":3,"encoding":"rgb8","queue_size":1,"tcp_no_delay":false,"payload_bytes":6220800,"mean_ns":1227622.784800,"median_ns":1202900.000000,"p95_ns":1376580.000000,"min_ns":991740.000000,"max_ns":4983630.000000,"stddev_ns":93504.184201,"throughput_mib_s":4832.605961,"throughput_gb_s":5.067355}
```

I have strong confidence that our ROS1 implementation (at least on the case of large image messages) is now as fast, or FASTER than noetic roscpp.

PR #334, opened 2026-07-17
Remove several client panics and unwraps

JesseGuillory-CM:
## Description
There were several instances where the rosbridge client could panic and enter itself into an unrecoverable state. These changes simply remove the panics and unwraps in favor of something that can actually be recovered to keep the connection stable.

## Checklist
- [ ] Update CHANGELOG.md

### DIFF
--- a/roslibrust_rosbridge/src/client.rs
+++ b/roslibrust_rosbridge/src/client.rs
@@ -358,18 +358,29 @@ impl ClientHandle {
         }
         {
             let mut comm = client.writer.write().await;
-            timeout(
+            if let Err(e) = timeout(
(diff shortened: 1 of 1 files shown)


Carter12s:
Thank you for the contribution!

Great to see the error handling improved!

I did a quick review and see no obvious issues or improvements with the changes. Should be able to get this merged shortly.

Carter12s:
While digging into this I spotted a potential deadlock where a disconnection while a service call was hanging could freeze up the client.

Fixing that ended up being a little messy, and I wanted to add some specific tests for these changes which was a little messy. Ended up pushing these to #335.

PR #338, opened 2026-08-30
Upgrade roslibrust_transforms to transforms v2.1

deniz-hofmeister:
## Description

Upgrades `roslibrust_transforms` from transforms v1.2 to v2.1 and adapts to its new API.

### Breaking changes

- Static transforms are now represented by `Stamp::Static` instead of a magic zero timestamp. On the wire they are still stamped with time zero.
- Transforms are built with `Transform::new(...)` / `Transform::static_between(...)` instead of struct literals, and fields are accessed via methods (`translation()`, `rotation()`, `parent()`, `child()`, `timestamp()`). Construction is fallible, so non-unit quaternions and NaN/infinite components are caught up front.
- `update_static_transform()` is removed. `add_transform()` now routes by the transform's stamp: `Stamp::At` publishes to `/tf`, `Stamp::Static` publishes to `/tf_static`. The registry is updated before publishing, so a transform the registry rejects (e.g. one that would re-parent a frame) is never sent to other nodes; rejection surfaces as a new `TransformManagerError::RejectedTransform`.
- `IntoTransform::into_transform` now returns `Result<Transform, TransformError>`.
- `Stamp`, `TimeError`, `TimePoint`, and `TransformError` are re-exported alongside the existing types.

### New features

- `latest_common_time(target, source)` — the tf2-style "latest available transform" lookup, returning `Stamp::At(time)` or `Stamp::Static` for purely static chains.
- `remove_frame(frame)` — drops a frame from the local buffer, providing an escape hatch for re-parenting (the transforms crate rejects a child appearing under a new parent otherwise).

### Robustness fixes

- Invalid or registry-rejected transforms received over `/tf` / `/tf_static` are dropped with a warning instead of silently corrupting the buffer (v1.2's infallible `add_transform` made this impossible to detect).
- Pre-epoch ROS timestamps (negative `sec`) are clamped to zero rather than wrapping, since `Timestamp` is unsigned nanoseconds.
- Waiter notification switched from a `broadcast` channel to a `watch` channel, so notifications coalesce and lagged-receiver handling disappears.
- Read-only lookups (`get_transform`, `get_transform_at`, `wait_for_transform` checks) take a read lock instead of a write lock, since v2 registry lookups no longer require `&mut`.

### Housekeeping

- Dropped the now-unused `chrono` dependency.
- Updated the examples and README to the new API.
- New tests: pre-epoch clamping, invalid transforms dropped, static publish routing to `/tf_static`, `latest_common_time`, and re-parenting rejected until `remove_frame` is called.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,11 +9,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- roslibrust_transforms gained `latest_common_time()` for looking up the newest time a transform can be served, and `remove_frame()` for removing a frame from the local buffer (e.g. to allow re-parenting it).
+
 ### Fixed
 
 - @JesseGuillory-CM removed several panics and poor error handling from rosbridge client.
 
 ### Changed
+
+- Upgraded roslibrust_transforms to transforms v2.1. Static transforms are now represented by `Stamp::Static` instead of a zero timestamp, transforms are built with `Transform::new` / `Transform::static_between` instead of struct literals, and `add_transform()` publishes static transforms to /tf_static automatically, replacing `update_static_transform()`. Invalid or conflicting transforms received over the wire are now dropped with a warning instead of silently corrupting the buffer.
 
 ## 0.21.0 - May 19th, 2026
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1525,6 +1525,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "hashbrown"
+version = "0.17.1"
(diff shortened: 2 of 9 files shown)


deniz-hofmeister:
Hey @Carter12s ,

I recently released a v2 of my transforms library, so I would like to propose this PR to update your transforms version to v2. Please have a look. 

Things that improved over v1 is an API overhaul with greatly improved Error reporting.

Carter12s:
Nice! Sorry for delay in processing this, but these changes look fantastic.

Thanks for the updates to transforms and for up-streaming them here!

PR #339, opened 2026-09-10
Make boolean literal parsing case-insensitive

althea28:
## Description
Make boolean literal parsing case-insensitive to match official rosidl_adapter.

Currently, roslibrust_codegen cannot parse .msg and .action files that contain capitalized boolean literals (e.g., bool success True). This causes serde_json::from_str to panic with "expected value at line XX column XX".

The official ROS 2 Python parser ([rosidl_adapter/parser.py](https://github.com/ros2/rosidl/blob/61aebc6362f495e377ac27178343c8f074e7eaf5/rosidl_adapter/rosidl_adapter/parser.py#L748)) parses boolean constants case-insensitively by converting the string to lowercase first: `if value_string.lower() not in (true_values + false_values):`

Because of this, several official ROS 2 packages (such as nav2_msgs in the navigation2 stack) currently use True or False in their definitions, which compile fine in standard ROS 2 but break roslibrust's strict generation. This PR adds a .to_lowercase() call when matching "bool" in parse_ros_value.

## Fixes
N/A

## Checklist
- [x] Update CHANGELOG.md (I did not update this as this change does not seem notable enough to include 😅 let me know if I should do this anyway!)


Carter12s:
Hey thanks for this contribution! Agreed, this is the appropriate fix and I really appreciate you running it down!

Will get a version with this fix out shortly!

PR #343, opened 2026-09-18
Add base64 deserialiser

althea28:
## Description
In ROS 2 and rosbridge, fixed-size byte arrays like uint8[16], commonly used in UUIDs, are optimized over the network by being encoded as Base64 strings. However, roslibrust_codegen strictly translates uint8[16] into a standard Rust [u8; 16]. When standard serde encounters this Base64 string from rosbridge, it expects a sequence of 16 integers and fails with a type mismatch, dropping the message. The error thrown is: `invalid type: string "...", expected an array of length 16` (where ... is the actual Base64 string payload)

This PR introduces Base64 compatibility for fixed-size arrays by:
- Adding fixed_serialize and fixed_deserialize in serde_rosmsg_bytes.rs using const generics (<const N: usize>) to handle byte arrays of any fixed length. The custom Visitor utilizes STANDARD.decode_slice to parse the Base64 string directly into a stack-allocated buffer, completely avoiding intermediate String and Vec<u8> heap allocations.
- Updating roslibrust_codegen/src/gen.rs so that when options.roslibrust_serde is enabled, the generator injects #[serde(serialize_with = "...", deserialize_with = "...")] onto any fixed-length uint8 array field, including those >32 elements.

## Fixes
Closes: N/A

## Checklist
- [/] Update CHANGELOG.md


Carter12s:
Gotcha. Placing some breadcrumbs here for myself:

This issue is a subset of: https://github.com/RosLibRust/roslibrust/issues/278 in that issue we fixed Vec<> but we didn't catch that the issue also affected fixed sized arrays.

I'll take a look at the solution in more detail tomorrow. At first glance this seems like a rougher way of solving the problem than we did in #279, and ideally we'd be able to work out a uniform solution.

Carter12s:
Ended up doing a bigger set of changes around this, but ultimately the concept here was spot on.

Thank you for finding the issue, flagging and raising the MR!

0.24 will have this fix and I'm expecting to release it this week.

